PR #1, opened 2019-06-26
Rewrite API, add unit tests

pbylicki:
First step of the rewrite. There are still things to be done, but it is already getting large.
Moved all logic related to serving and presenting documentation data to FastAPI based web app
There are is still some logic to be added for search endpoint, I will add it in the meantime.
Added unit tests.
Not added Robot tests, I guess that old ones will require some extension, probably using DatabaseLibrary keywords
Packaging and dockerization is not there, but it should be pretty similar to what we did in old project.


MaciejWiczk (approved):
Awesome!

Issue #6, opened 2019-06-29
Add endpoints for updating data to API


pbylicki:
We need to introduce endpoints for updating data via REST API. Required endpoints are:
- add collection
- add keyword
- update collection
- update keyword
- delete collection
- delete keyword

Scope of this task should not include authentication/authorization, it will be covered as separate task.

MaciejWiczk:
Done in #13

Issue #3, opened 2019-06-29
Migrate and extend robot tests from RfHub2


MaciejWiczk:
Migrate and extend robot tests from RfHub2. Tests should use chrome and selenium and verify running of application, UI and data correctness. All tests should be added to TravisCI pipeline.

Issue #2, opened 2019-06-29
Add functionality to populate RfHub-new with data


MaciejWiczk:
This task is about adding functionality to populate RfHub-new as in previous application. There is a need of checking specified folders, with keywords and libraries, which should be added and displayed in application. Adding collections and keywords should be done via API.

pbylicki:
IMO it should be Click CLI application (we already depend on this package) located in rfhub2.cli package structured in such a way that:
- client responsible for communication with REST API (or database if needed before #6 is implemented) should be defined, ideally with one interface that covers both ways of communication (REST/db)
- it would be nice to use the same model for sending data that API uses as incoming data model for parsing and validation (it will be implemented as part of #6).

MaciejWiczk:
I've started looking at Click package. I need some more understanding in regards of making clients for Rest API/DB. For now I know that populating application:
- using DB should happen before app starts
- using Rest API should happen after app starts, so we have working endpoint, obviously

Should those two clients be located in rfhub2.cli? Or maybe somewhere different like rfhub2.clients? Last but not least, should app.py with create_app method be changed into class and incorporate click args, and various startup modes (ie. worked or web, DB or Rest population)

pbylicki:
Once #13 is merged I do not see a point in updating data directly in db, we should use Rest API and assume that web application will be run separately from the job populating the data. With this approach rest client should be just part of the `rfhub2.cli` package. I would make cli only about populating the data, not running the web server so let's leave web server launching code as is for this task.
So in `rfhub2.cli` there is an Click application for populating rfhub2 web server running on provided url that should be run by `python -m rfhub2.cli [any args here]`
Do you agree @MaciejWiczk ?

MaciejWiczk:
Ok, now I have better picture. I see two actions to be taken care of:
- delete row: https://github.com/pbylicki/rfhub-new/blob/master/rfhub2/main.py#. @pbylicki will this lead to change in unit_tests? Or are they independent of that?
- creation of another image with rfhub.cli, because docker endpoint will be different han for whole app.

pbylicki:
The unit tests use the same function, but independently in their setup, so recreating data can be removed from application starting procedure once mechanism for populating data is done.
[...]

PR #9, opened 2019-06-30
Adding CodeCov to pipeline

MaciejWiczk:
This is a first step to monitor application coverage. Appropriate requirements and badges were added. After #6, #2 and #3 I would like to try to use coverage from robot tests as well.


pbylicki (approved):
Thanks, it looks awesome!

PR #11, opened 2019-07-05
Adding docker images

MaciejWiczk:
Dockerfiles for both SQLite an PostgreSQL images

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ docker run -it -p 8000:8000 rfhub-new
 ```
 To run application using docker image with Postgres database
 ```
-docker run -it -p 8000:8000 -e RFHUB_DB_URI="postgresql://postgres:postgres@localhost:5432/postgres" rfhub-new-postgres
+docker run -it -p 8000:8000 --network=host -e RFHUB_DB_URI="postgresql://postgres:postgres@localhost:5432/postgres" rfhub-new-postgres
 ```
 #### Run unit tests
 ```
--- a/README.md
+++ b/README.md
@@ -8,24 +8,36 @@
 ```
 pip install -r requirements.txt -r requirements-dev.txt
 ```
-
 To run app with Postgres db install additional dependencies:
 ```
 pip install -r requirements-postgres.txt
 ```
-
+#### Build docker image
+To build image using SQLite DB
+```
+docker build -f docker/Dockerfile -t rfhub-new .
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -8,24 +8,36 @@
...
 RFHUB_DB_URI=postgresql://postgres:postgres@localhost:5432/postgres python -m rfhub2
 ```
-
+To run application using docker image with default (SQLite) database
+```
+docker run -it -p 8000:8000 rfhub-new
+```
+To run application using docker image with Postgres database
+```
+docker run -it -p 8000:8000 -e RFHUB_DB_URI="postgresql://postgres:postgres@localhost:5432/postgres" rfhub-new-postgres
```
pbylicki: This command may be misleading because it expects postgres to be available via localhost interface of the rfhub container. It can be changed either by switching localhost to docker network ip of postgres container or to run rfhub with `--network=host` option. I would go with the second option here, since it can be copied and run without any adjustments.

docker/Dockerfile-postgres
```diff
@@ -0,0 +1,22 @@
+FROM python:3.6-alpine
+
+COPY requirements*.txt /tmp/
+
+RUN apk update && apk upgrade
```
pbylicki: this should not be necessary

docker/Dockerfile-postgres
```diff
@@ -0,0 +1,22 @@
...
+
+RUN apk add --no-cache --virtual .build-deps gcc musl-dev make postgresql-dev && \
+    apk add --update --no-cache postgresql &&\
```
pbylicki: this command installs more than we need, we can change it to just `apk add --no-cache libpq`

docker/Dockerfile-postgres
```diff
@@ -0,0 +1,22 @@
...
+    pip install --disable-pip-version-check --no-cache-dir -r /tmp/requirements.txt \
+    -r /tmp/requirements-postgres.txt && \
+    apk del .build-deps && \
+    rm /tmp/requirements*.txt && \
+    mkdir /app && \
+    chown 1001:1001 /app
```
pbylicki: this ownership setup has no apparent purpose for this image, but let's maybe keep it for now and for the sake of uniformity

pbylicki (approved):

PR #13, opened 2019-07-06
Add API endpoints for updating data

pbylicki:
Implements #6

### DIFF
--- a/rfhub2/api/endpoints/collections.py
+++ b/rfhub2/api/endpoints/collections.py
@@ -1,10 +1,12 @@
-from fastapi import APIRouter, Depends
+from fastapi import APIRouter, Depends, HTTPException
+from starlette.responses import Response
 from typing import List, Optional
 
 from rfhub2.api.utils.db import get_collection_repository
 from rfhub2.api.utils.http import or_404
+from rfhub2.db.base import Collection as DBCollection
 from rfhub2.db.repository.collection_repository import CollectionRepository
-from rfhub2.model import Collection
+from rfhub2.model import Collection, CollectionUpdate
 
 router = APIRouter()
 
@@ -24,3 +26,29 @@ def get_collection(*, repository: CollectionRepository = Depends(get_collection_
                    id: int):
     collection: Optional[Collection] = repository.get(id)
     return or_404(collection)
+
+
+@router.post("/", response_model=Collection, status_code=201)
+def create_collection(*, repository: CollectionRepository = Depends(get_collection_repository),
+                      collection: CollectionUpdate):
+    db_collection = DBCollection(**collection.dict())
+    return repository.add(db_collection)
+
+
+@router.put("/{id}/", response_model=Collection)
+def update_collection(*, repository: CollectionRepository = Depends(get_collection_repository),
+                      id: int,
+                      collection_update: CollectionUpdate):
+    db_collection: DBCollection = or_404(repository.get(id))
+    updated: DBCollection = repository.update(db_collection, collection_update.dict(skip_defaults=True))
+    return updated
+
+
+@router.delete("/{id}/")
+def delete_collection(*, repository: CollectionRepository = Depends(get_collection_repository),
+                      id: int):
+    deleted: int = repository.delete(id)
+    if deleted:
+        return Response(status_code=204)
+    else:
+        raise HTTPException(status_code=404)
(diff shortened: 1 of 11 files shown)


rfhub2/db/repository/keyword_repository.py
```diff
@@ -24,6 +25,11 @@ def add(self, keyword: Keyword) -> Keyword:
         self.session.refresh(keyword)
         return keyword
 
+    def delete(self, keyword_id: int) -> int:
+        row_count = self._keywords.filter(Keyword.id == keyword_id).delete()
+        self.session.commit()
+        return row_count
```
MaciejWiczk: Why return row_count? Shouldn't it be single keywords with such id? From this method I assume that each keyword has unique id in general, not only in single collection?

rfhub2/model/__init__.py
```diff
@@ -10,8 +10,7 @@ class Config:
         orm_mode = True
 
 
-class NestedKeyword(BaseModel):
-    id: int
+class KeywordUpdate(BaseModel):
```
MaciejWiczk, line 13: Why update method does not need id?

rfhub2/db/repository/keyword_repository.py
pbylicki: This way we can say if any row was deleted. If you will see the delete endpoint that uses this method, this way we can inform user that keyword with this id was properly removed if the row_count is 1 and respond with 404 if it was 0 to show that there was no record with this id to be removed. As you wrote, the keys are unique so this value should be 0/1

rfhub2/model/__init__.py
pbylicki: This model is basically set of attributes that should be updateable on keyword record. Id is already passed in endpoint url so by removing it from this model we can easier update the record without any special consideration for attributes that should not be updateable.
It does not mean, that request body cannot contain attribute id, it just means that it won't be considered as updateable field so we do not need to care if someone wants to overwrite it via this request.

rfhub2/api/endpoints/keywords.py
```diff
@@ -24,3 +27,33 @@ def get_keyword(*, repository: KeywordRepository = Depends(get_keyword_repositor
...
+def update_keyword(*, repository: KeywordRepository = Depends(get_keyword_repository),
+                   id: int,
+                   keyword_update: KeywordUpdate):
+    db_keyword: DBKeyword = or_404(repository.get(id))
+    updated: DBKeyword = repository.update(db_keyword, keyword_update.dict(skip_defaults=True))
+    return updated
+
+
+@router.delete("/{id}/")
+def delete_collection(*, repository: KeywordRepository = Depends(get_keyword_repository),
```
MaciejWiczk: delete_keyword?

pbylicki: yes, careless copying

rfhub2/api/endpoints/collections.py
```diff
@@ -24,3 +26,29 @@ def get_collection(*, repository: CollectionRepository = Depends(get_collection_
...
+                      collection_update: CollectionUpdate):
+    db_collection: DBCollection = or_404(repository.get(id))
+    updated: DBCollection = repository.update(db_collection, collection_update.dict(skip_defaults=True))
+    return updated
+
+
+@router.delete("/{id}/")
+def delete_collection(*, repository: CollectionRepository = Depends(get_collection_repository),
+                      id: int):
+    deleted: int = repository.delete(id)
```
MaciejWiczk, line 50: I can see in code that this method is deleting collection https://github.com/pbylicki/rfhub-new/blob/master/rfhub2/db/repository/collection_repository.py#L18 using ```return self.session.query(Collection).options(selectinload(Collection.keywords))```. Does this do cascade delete so that keywords are dropped first, and then collection itself?

rfhub2/model/__init__.py
MaciejWiczk: Ok, seems clever. I'm still trying to understand FastAPI framework.

rfhub2/api/endpoints/collections.py
MaciejWiczk: Ok, from SQLAlchemy docs I see it does cascade delete. Splenind!

pbylicki: Yes, it does, but it is worth remembering that in order to make it work on SQLite, you need this:
https://github.com/pbylicki/rfhub-new/blob/master/rfhub2/db/session.py#L18

rfhub2/api/endpoints/keywords.py
pbylicki: fixed

pbylicki:
@MaciejWiczk I have made additional refactor of db repository code

MaciejWiczk (approved):

PR #16, opened 2019-07-08
Add Basic auth support

pbylicki:
Add obligatory basic auth to all writing API endpoints. Currently it supports just one authorized user defined in config via environment variables.


MaciejWiczk (approved):
Splendid!

PR #20, opened 2019-07-19
Feature/rfhub new 2 add functionality to populate app

MaciejWiczk:
This is a very, very early stage of changes for rfhub-new.cli. @pbylicki I would be glad to receive feedback and review. 

- Some methods were taken "as is" from rfhub2/rfhub, just to check if everything is working as expected/as used to
- Type hints needs more refinement, I can manage simple one, but more sophisticated are a mystery to me
- Unit tests are failing, and I didn't add any, because logic may change a lot from current state
- Mechanism to skip library addition if such already exists would be nice feature, but this is not yet time for it


rfhub2/cli/population.py
```diff
@@ -0,0 +1,175 @@
+# from pathlib import Path
+import robot.libraries
+from robot.errors import DataError
+from robot.libdocpkg import LibraryDocumentation
+from requests import session, post
+from rfhub2.config import APP_INTERFACE, APP_PORT, BASIC_AUTH_USER, BASIC_AUTH_PASSWORD
```
pbylicki: Let's stick to convention where you have first group of external imports in alphabetical order and then, after one empty line internal, rfhub imports

rfhub2/cli/population.py
```diff
@@ -0,0 +1,175 @@
...
+import os
+import re
+
+RESOURCE_PATTERNS = {".robot", ".txt", ".tsv", ".resource"}
+ALL_PATTERNS = (RESOURCE_PATTERNS | {".xml", ".py"})
+PROTOCOL = 'http://'
```
pbylicki: it shouldn't be hardcoded, should be part of user provided api url instead

rfhub2/cli/population.py
```diff
@@ -0,0 +1,175 @@
...
+API_V1 = 'api/v1'
+
+
+class LibraryPopulation(object):
```
pbylicki: maybe we can rename it, for example RfhubImporter or RfhubClient?

rfhub2/cli/population.py
```diff
@@ -0,0 +1,175 @@
...
+                    if re.match(r'Test Cases?', match.group(1), re.IGNORECASE):
+                        # if there's a test case table, it's not a keyword file
+                        return False
+
+                    if (not found_keyword_table and
+                            re.match(r'(User )?Keywords?', match.group(1), re.IGNORECASE)):
+                        found_keyword_table = True
+        return found_keyword_table
+
+    def _post_request(self, session, endpoint, data):
```
pbylicki: all http communication should be moved to separate class, like ApiClient with methods allowing for adding/deleting keywords and collections. These methods should ideally take objects of these classes as arguments: https://github.com/pbylicki/rfhub-new/blob/master/rfhub2/model/__init__.py based on the expected input model in api endpoint that they are calling.

rfhub2/cli/population.py
```diff
@@ -0,0 +1,175 @@
...
+        self.auth = (BASIC_AUTH_USER, BASIC_AUTH_PASSWORD)
+        self.no_installed_keywords = no_installed_keywords
+
+    def add_collections(self) -> None:
+        """
+        Traverses through paths and adds libraries to rfhub.
+        :return:
+        """
+
+        def traverse_paths(path):
```
pbylicki: I think the function does not need to be nested, you can add `_` to indicate that it is private. Similar note to other nested methods below. This way you will be able to unit test them

rfhub2/cli/population.py
```diff
@@ -0,0 +1,175 @@
...
+            for item in os.listdir(path):
```
pbylicki: Maybe you could switch to using `pathlib` module in general?

rfhub2/cli/population.py
```diff
@@ -0,0 +1,175 @@
...
+                lib_dict['doc_format'] = lib_dict.pop('_setter__doc_format')
+                for key in ('_setter__keywords', 'inits', 'named_args'):
+                    lib_dict.pop(key)
+                return lib_dict
+
+            lib_dict = _libdoc_to_dict()
+            lib_dict['path'] = path
+            return lib_dict
+
+        def _serialise_keywords():
```
pbylicki: these method should map to pydantic-based keyword model

rfhub2/cli/population.py
```diff
@@ -0,0 +1,175 @@
...
+            keywords = [keyword.__dict__ for keyword in libdoc.keywords]
+            for keyword in keywords:
+                keyword.pop('tags')
```
pbylicki: Maybe we should add keyword tags into our model, what do you think?

rfhub2/cli/population.py
```diff
@@ -0,0 +1,175 @@
...
+                    keyword["args"] = str([str(item) for item in keyword["args"]]).replace("\'", "\"")
+                else:
+                    keyword["args"] = ""
+            return keywords
+
+        libdoc = LibraryDocumentation(path)
+        serialised_keywords = _serialise_keywords()
+        serialised_libdoc = _serialise_libdoc()
+        s = session()
+        coll_req = self._post_request(s, 'collections', serialised_libdoc)
```
pbylicki: I think it would be nice if it would be refactor in such a way that ApiClient method would both take as an argument and return instances of model classes, so you would not have any request/json related logic here

rfhub2/cli/population.py
MaciejWiczk: moved as a part of app-protocol

rfhub2/cli/population.py
MaciejWiczk: renamed to RfhubImporter

rfhub2/cli/population.py
MaciejWiczk: First part done, second part needs review

rfhub2/cli/population.py
MaciejWiczk: AFAIK pathlib does not handle listing directories and path.walk.

rfhub2/cli/population.py
MaciejWiczk: That would be nice :)

rfhub2/cli/population.py
MaciejWiczk: Extracted and unnested. Valid point with tests

rfhub2/cli/population.py
pbylicki: Valid point with walk. Please see `os.scandir` or `Path.iterdir` as substitute for `os.listdir`. `os` functions work with Path objects so you can switch to pathlib and use os functions when necessary. Most use cases in our code should be replaceable and result in more concise code

rfhub2/cli/population.py
MaciejWiczk: Changed parameter to app-url, default value is `http://localhost:8000`

rfhub2/cli/population.py
MaciejWiczk: Can You take a look, at changes I did? I'm still not sure about it.

rfhub2/cli/rfhub_importer.py
```diff
@@ -109,27 +109,27 @@ def _serialise_keywords(self, libdoc) -> Keyword:
         return keywords
 
     @staticmethod
-    def _is_library_with_init(path: str) -> bool:
-        return os.path.isfile(os.path.join(path, '__init__.py')) and \
-            len(LibraryDocumentation(path).keywords) > 0
+    def _is_library_with_init(path: Path) -> bool:
+        return Path.is_file(path.joinpath('__init__.py')) and \
```
pbylicki: `(path  / '__init__.py').is_file()`

rfhub2/cli/rfhub_importer.py
```diff
@@ -39,34 +40,33 @@ def add_collections(self) -> None:
...
         """
         Traverses through paths and adds libraries to rfhub.
         Helper function for add_collections.
         """
-        for item in os.listdir(path):
-            full_path = os.path.join(path, item)
-            if os.path.isdir(full_path):
-                if self._is_library_with_init(full_path):
-                    self.add(full_path)
+        for item in Path.iterdir(Path(path)):
```
pbylicki: path and item are already instances of Path, can be just path.iterdir(), item.is_dir(), item.is_file() etc

rfhub2/cli/rfhub_importer.py
```diff
@@ -109,27 +109,27 @@ def _serialise_keywords(self, libdoc) -> Keyword:
...
     @staticmethod
-    def _is_libdoc_file(file: str) -> bool:
+    def _is_libdoc_file(file: Path) -> bool:
         """Return true if an xml file looks like a libdoc file"""
         # inefficient since we end up reading the file twice,
         # but it's fast enough for our purposes, and prevents
         # us from doing a full parse of files that are obviously
         # not libdoc files
-        if file.lower().endswith(".xml"):
+        if file.suffix == '.xml': #Path(file).suffix == ".xml":
```
pbylicki: unnecessary comment

rfhub2/cli/rfhub_importer.py
MaciejWiczk: Done

rfhub2/cli/rfhub_importer.py
MaciejWiczk: Changed. Turned out that path was passed as string instead of Path object. Now it's cast as Path.

rfhub2/cli/rfhub_importer.py
MaciejWiczk: Yep, my bad. Fixed.

rfhub2/cli/population.py
MaciejWiczk: Replaced os.listdir with path.iterdir()

rfhub2/cli/api_client.py
```diff
@@ -13,7 +13,7 @@ class Client(object):
     """
     API client with methods to populate rfhub2 application.
     """
-    def __init__(self, app_url: int, user: str, password: str,):
+    def __init__(self, app_url: str, user: str, password: str,):
```
pbylicki: unnecessary comma at the end

tests/cli/rfhub_importer.py
```diff
@@ -0,0 +1,118 @@
...
+from pathlib import Path
+
+from rfhub2.cli.rfhub_importer import RfhubImporter
+from rfhub2.cli.api_client import Client
+
+
+class RfhubImporterTests(unittest.TestCase):
+
+    def setUp(self) -> None:
+        self.fixture_path = Path.cwd().joinpath('tests', 'acceptance', 'fixtures')
```
pbylicki: maybe you could switch to using `/` operator instead of `joinpath`?

tests/cli/rfhub_importer.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def test_is_library_with_init_should_return_true_on_library_with_init(self):
+        file = self.fixture_path.joinpath('LibWithInit')
+        result = self.rfhub_importer._is_library_with_init(file)
+        self.assertTrue(result, 'method should return true if file is python library with init')
+
+    def test_is_library_with_init_should_return_false_on_library_without_init(self):
+        file = self.fixture_path.joinpath('LibsWithEmptyInit')
+        result = self.rfhub_importer._is_library_with_init(file)
+        self.assertFalse(result, 'method should return true if file is python library without init')
```
pbylicki: the message should be about false case

tests/cli/rfhub_importer.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def test_is_robot_keyword_file_should_return_true_on_libdoc(self):
+        file = self.fixture_path.joinpath('test_libdoc_file.xml')
+        result = self.rfhub_importer._is_robot_keyword_file(file)
+        self.assertTrue(result, 'method should return true if file is libdoc file')
+
+    def test_is_library_file_should_return_true_on_resource(self):
+        file = self.fixture_path.joinpath('test_resource.resource')
+        result = self.rfhub_importer._is_robot_keyword_file(file)
+        self.assertTrue(result, 'method should return true if file is python library')
```
pbylicki: Message should be updated

tests/cli/rfhub_importer.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def test_is_library_file_should_return_false_on_lib_with_init(self):
+        file = self.fixture_path.joinpath('LibWithInit', '__init__.py')
+        result = RfhubImporter._is_library_file(file)
+        self.assertFalse(result, 'method should return true if file is python library')
```
pbylicki, line 189: Message should be updated

tests/cli/rfhub_importer.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def test_has_keyword_table_should_return_true(self):
+        data = '*** Keywords ***'
+        result = RfhubImporter._has_keyword_table(data=data)
+        self.assertTrue(result, 'method should return true if Keywords were found')
+
+    def test_has_keyword_table_should_return_false(self):
+        data = '*** Keys ***'
+        result = RfhubImporter._has_keyword_table(data=data)
+        self.assertFalse(result, 'method should return true if Keywords were found')
```
pbylicki: Message should be updated

tests/cli/rfhub_importer.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def test_has_test_case_table_should_return_true(self):
+        data = '*** Test Case ***'
+        result = RfhubImporter._has_test_case_table(data=data)
+        self.assertTrue(result, 'method should return true if Keywords were found')
```
pbylicki: Message should be updated

tests/cli/rfhub_importer.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def test_has_test_case_table_should_return_false(self):
+        data = '*** Test ***'
+        result = RfhubImporter._has_test_case_table(data=data)
+        self.assertFalse(result, 'method should return true if Keywords were found')
```
pbylicki: Message should be updated

rfhub2/cli/rfhub_importer.py
```diff
@@ -151,26 +151,26 @@ def _should_ignore(file: Path) -> bool:
...
-                        found_keyword_table = True
-        return found_keyword_table
+                return not RfhubImporter._has_test_case_table(data) and \
+                    RfhubImporter._has_keyword_table(data)
+        return False
+
+    @staticmethod
+    def _has_keyword_table(data: str) -> bool:
+        """Returns true if file has keyword or user keyword table"""
+        return len(re.findall(r'^\*+\s*((?:User )?Keywords?)',
```
pbylicki: I think using `re.search` would be more efficient since one occurence is enough to satisfy the condition

rfhub2/cli/rfhub_importer.py
```diff
@@ -151,26 +151,26 @@ def _should_ignore(file: Path) -> bool:
...
+                              data, re.MULTILINE | re.IGNORECASE)) > 0
+
+    @staticmethod
+    def _has_test_case_table(data: str) -> bool:
+        """Returns true if file has keyword or user keyword table"""
+        return len(re.findall(r'^\*+\s*(Test Cases?)', data, re.MULTILINE | re.IGNORECASE)) > 0
```
pbylicki: `re.search` as above

tests/cli/rfhub_importer.py
MaciejWiczk: yeah, together with test name to reflect used method.

rfhub2/cli/rfhub_importer.py
MaciejWiczk: Done

rfhub2/cli/rfhub_importer.py
MaciejWiczk: Done

tests/cli/rfhub_importer.py
MaciejWiczk: fixed

tests/cli/rfhub_importer.py
MaciejWiczk: fixed

rfhub2/cli/rfhub_importer.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+class RfhubImporter(object):
+
+    def __init__(self, paths: Tuple[Path, ...], no_installed_keywords: bool, client: Client) -> None:
+        self.paths = paths
+        self.no_installed_keywords = no_installed_keywords
+        self.client = client
+
+    def delete_collections(self) -> None:
```
MaciejWiczk: @pbylicki I do lack of idea how to unit test that method, as well as `add_collections`, `_traverse_paths` and `add`. Could You provide some ideas? I really want to keep high coverage :)

rfhub2/cli/api_client.py
```diff
@@ -0,0 +1,79 @@
...
+                  f'Check parameters for app-interace and port.')
+            exit(1)
+        req_check_auth = self._post_request('collections', TEST_COLLECTION)
+        if req_check_auth['name'] == 'healtcheck_collection':
+            self._delete_request('collections', req_check_auth["id"])
+        else:
+            print(f'{req_check_auth.reason}! Check used credentials.')
+            exit(1)
+
+    def get_collections(self) -> List[Collection]:
```
MaciejWiczk: @pbylicki Should I check this methods similar to this test:
https://github.com/pbylicki/rfhub-new/blob/master/tests/api/endpoints/healthcheck_tests.py#L17
Or should I start app similarly to BaseEnpointTests and connect to working app?

pbylicki: I would mock requests library methods (get, post, etc.) that are used to perform actual http request with different responses that you need for each scenario. You can either prepare the mocking by yourself or use some utility like this: https://github.com/getsentry/responses
I you want I can help you with some initial test as an example.

rfhub2/cli/rfhub_importer.py
pbylicki: @MaciejWiczk As for the http request parts, see my other comment on unit testing them. As for file system operations with pathlib, you could also mock the methods from this library to return expected objects. There are some nice tutorials on the internet how to use built-in unittest.mock package, but if you want I can prepare test as an example for you

rfhub2/cli/api_client.py
MaciejWiczk: > I you want I can help you with some initial test as an example.
 
That would be very good idea. I tried mockng and responses, and dependency override and got stuck.

I have smth like this:
```
    def test_get_collections(self):
        with responses.RequestsMock() as rsps:
            rsps.add(responses.GET, f'{self.app_url}/api/v1/collections', body=json.dumps(RESP),
                     status=200, content_type='application/json')
            resp = self.client.get_collections()
            self.assertEqual(resp, RESP)
```
Is it a good way?

pbylicki: Yes, this is what I meant. Note that in `rsps.add` you can just pass `json=RESP` to provide expected body.

rfhub2/cli/rfhub_importer.py
pbylicki: @MaciejWiczk As for testing methods relying on pathlib methods, here is simple example of how you could do it:

```
    def test_traverse_paths(self):
        def mock_iterdir():
            return [self.fixture_path / 'SingleClassLib' / 'SingleClassLib.py']
        path = Mock()
        path.iterdir = mock_iterdir
        self.rfhub_importer._traverse_paths(path)`
```
You can provide mocked iterdir method returning instances of Path objects matching conditions that you want to test, or if you do not have existing files you could also returned mocked Path object in that list.

Current implementation of this methods is not really easy for testing because they do not return anything, and the flow is not very straightforward. What I would do is to change the implementation of the flow to be something like this:
1. method to get sequence of valid library paths based on input parameters
2. method to map path to Collection object (containing list of Keyword objects)
3. method to send collection and its keywords using client (it can also return not None response for some final report of how many were sent etc. but it is not necessary in this PR)

In this way testing this three methods should be very easy, with method 1 relying on mocked iterdir like in example above, method 2 requiring no mocks and method 3 relying on mock of client's add collection/keyword methods. I think that struggling to create unit tests without  such refactor has no point.

rfhub2/cli/api_client.py
MaciejWiczk: That's great! I've aded some test for api_client.py. Coverage is 100% with more mocks than real code ;)

rfhub2/cli/rfhub_importer.py
MaciejWiczk: @pbylicki Ok, so I will start with rafactoring rfhub_importer.py according to points 1-3. regarding not None response, I do print out library name and how many keywords were added:
```
BuiltIn library with 104 keywords loaded.
Collections library with 43 keywords loaded.
DateTime library with 8 keywords loaded.
Easter library with 1 keywords loaded.
```
Do You have something else in mind?
PS. Regarding point 2, should collection objects be stored in memory and then sent to app? They are not big so I think it won't be a problem.

pbylicki: @MaciejWiczk I thought about displaying total number of loaded collections and keywords, but again, it does not need to be part of this PR.
Yes, I would not expect a scenario that collection objects would be too big to store in memory. You can also use yielding results from this method one by one instead of returning whole list if you want.

MaciejWiczk: @pbylicki one more question regarding point 2. Can collection with keywords be posted as one, or should keywords be posted separately as they are now?

pbylicki: API does not allow posting collection with nested keywords, but you can pass collection with keywords to method which then will send collection first, and all its keywords after that

MaciejWiczk: Ok, that's cool. I will do exactly as You say. I'm almost done with unit testing, hopefully it will work as I expect in e2e ;)

MaciejWiczk:
@pbylicki I think that PR is ready for another round of comments. I'm thinking about testing click.cli, but it would require to mock main method. Is it still reasonable to do that? I assume it would end up in returning specified parameters/flags. I see more sense to test it in e2e using robot, but I;m open for suggestions.

rfhub2/cli/api_client.py
```diff
@@ -0,0 +1,79 @@
...
+    API client with methods to populate rfhub2 application.
+    """
+    def __init__(self, app_url: str, user: str, password: str):
+        self.app_url = app_url
+        self.session = session()
+        self.api_url = f'{self.app_url}/{API_V1}'
+        self.session.auth = (user, password)
+        self.session.headers = {"Content-Type": "application/json", "accept": "application/json"}
+
+    def check_communication_with_app(self) -> None:
```
pbylicki: I think that my comment got lost in one of the previous PRs. I think that this check is too much, with adding and deleting test data on each run. For me it is enough that application will raise exception about user being unauthenticated on first post request. The meaningful error message at that point will give user the same information without any additional logic and any significant waiting time. If you want we can keep basic connectivity check here - for this i would suggest using healthcheck endpoint instead.

rfhub2/cli/api_client.py
```diff
@@ -0,0 +1,79 @@
...
+                  f'Check parameter for app_url.')
+            exit(1)
+        req_check_auth = self._post_request('collections', TEST_COLLECTION)
+        if req_check_auth['name'] == 'healtcheck_collection':
+            self._delete_request('collections', req_check_auth["id"])
+        else:
+            print('Check used credentials!')
+            exit(1)
+
+    def get_collections(self) -> List[Collection]:
```
pbylicki: The type annotations here are misleading and AFAIR are displayed as warnings in PyCharm. Until we move to actually use here pydantic models, they should be changed to `Dict`

rfhub2/cli/cli.py
```diff
@@ -0,0 +1,31 @@
...
+              help="Specifies rfhub2 user to authenticate on endpoints that requires that. Default value is rfhub.")
+@click.option('-p', '--password', type=click.STRING, default="rfhub",
+              help="Specifies rfhub2 password to authenticate on endpoints that requires that. Default value is rfhub.")
+@click.option('--no-installed-keywords', type=click.BOOL, default=False, is_flag=True,
+              help='Flag specifying if package should skip loading commonly installed libraries, '
+                   'such as such as BuiltIn, Collections, DateTime etc.')
+@click.option('--no-db-flush', type=click.BOOL, default=False, is_flag=True,
+              help='Flag specifying if package should delete from rfhub2 all existing libraries.')
+@click.argument('--paths', nargs=-1, type=click.Path(exists=True))
+def main(app_url: str, user: str, password: str, __paths: Tuple[Path, ...],
```
pbylicki: why `__paths` and not just `paths` ?

rfhub2/cli/rfhub_importer.py
```diff
@@ -0,0 +1,216 @@
...
+RESOURCE_PATTERNS = {'.robot', '.txt', '.tsv', '.resource'}
+ALL_PATTERNS = (RESOURCE_PATTERNS | {'.xml', '.py'})
+EXCLUDED_LIBRARIES = {'remote.py', 'reserved.py', 'dialogs.py', 'dialogs_jy.py',
+                      'dialogs_py.py', 'dialogs_ipy.py', 'setup.py'}
+INIT_FILES = {'__init__.txt', '__init__.robot', '__init__.html', '__init__.tsv'}
+
+
+class RfhubImporter(object):
+
+    def __init__(self, paths: Tuple[Path, ...], no_installed_keywords: bool, client: Client) -> None:
```
pbylicki: I would reorder the params to make client first. You usually put this kind of flags like `no_installed_keywords` at the end of argument list

rfhub2/cli/rfhub_importer.py
```diff
@@ -0,0 +1,216 @@
...
+        self.paths = paths
+        self.no_installed_keywords = no_installed_keywords
+        self.client = client
+
+    def delete_collections(self) -> Set[int]:
+        """
+        Deletes all existing collections.
+        """
+        self.client.check_communication_with_app()
+        collections = self.client.get_collections()
```
pbylicki, line 29: it is worth to remember about the possible multiple pages of results here, but I think we can skip it for now.

rfhub2/cli/rfhub_importer.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def import_libraries(self) -> Tuple[int, int]:
+        """
+        Import libraries to application from paths specified when invoking client.
+        :return: Number of libraries loaded
+        """
+        libraries_paths = self.get_libraries_paths()
+        collections = self.create_collections(libraries_paths)
+        loaded_collections = self.add_collections(collections)
+        return len(loaded_collections), sum(v for dict in loaded_collections for v in dict.values())
```
pbylicki: you shadow `dict` class here. it is only local scope and it does not affect anything, but still it is rather bad convention.

rfhub2/cli/rfhub_importer.py
```diff
@@ -0,0 +1,216 @@
...
+        return [self.create_collection(path) for path in paths]
+        # ToDo try except to handle problem with library creation
+
+    def create_collection(self, path: Path) -> Collection:
+        """
+        Creates Collection object from provided path.
+        :param path: Path
+        :return: Collection object
+        """
+        libdoc = LibraryDocumentation(str(path))
```
pbylicki, line 100: is conversion to string necessary here?

rfhub2/cli/rfhub_importer.py
```diff
@@ -0,0 +1,216 @@
...
+        """
+        loaded_collections = []
+        for collection in collections:
+            coll_req = self.client.add_collection(collection)
+            if coll_req['name'] == collection['name']:
+                collection_id = coll_req['id']
+                for keyword in collection['keywords']:
+                    keyword['collection_id'] = collection_id
+                    self.client.add_keyword(keyword)
+                loaded_collections.append({collection['name']: len(collection["keywords"])})
```
pbylicki: How about changing the dict to: `{'name': collection['name'], 'keywords': len(collection["keywords"])}` to have more object-like dicts with well defined keys?

rfhub2/cli/rfhub_importer.py
```diff
@@ -0,0 +1,216 @@
...
+    def add_collections(self, collections: List[Collection]) -> List[Dict[str, int]]:
+        """
+        Adds collections and keywords from provided list to app.
+        :param collections: List of collections object
+        :return: list of dictionaries with collection name and number of keywords.
+        """
+        loaded_collections = []
+        for collection in collections:
+            coll_req = self.client.add_collection(collection)
+            if coll_req['name'] == collection['name']:
```
pbylicki: What is the scenario to make this condition fail without throwing an exception? it seems to me that if anything goes wrong with adding the collection, either `add_collection` will throw exception or `coll_req` won't have `name` key and KeyError will be thrown.

rfhub2/cli/rfhub_importer.py
```diff
@@ -0,0 +1,216 @@
...
+                for keyword in collection['keywords']:
+                    keyword['collection_id'] = collection_id
+                    self.client.add_keyword(keyword)
+                loaded_collections.append({collection['name']: len(collection["keywords"])})
+                print(f'{collection["name"]} library with {len(collection["keywords"])} keywords loaded.')
+            else:
+                print(f'{collection["name"]} library was not loaded!')
+        return loaded_collections
+
+    def _serialise_libdoc(self, libdoc: LibraryDocumentation, path: str, keywords: List[Keyword]) -> Collection:
```
pbylicki: should be dict in annotations

rfhub2/cli/rfhub_importer.py
```diff
@@ -0,0 +1,216 @@
...
+
+        lib_dict = libdoc.__dict__
+        lib_dict['doc_format'] = lib_dict.pop('_setter__doc_format')
+        for key in ('_setter__keywords', 'inits', 'named_args'):
+            lib_dict.pop(key)
+        lib_dict['path'] = path
+        lib_dict['keywords'] = keywords
+        return lib_dict
+
+    def _serialise_keywords(self, libdoc: LibraryDocumentation) -> Keyword:
```
pbylicki: should be dict in annotations

tests/cli/rfhub_importer.py
```diff
@@ -0,0 +1,296 @@
...
+            result = self.rfhub_importer.delete_collections()
+            self.assertEqual({2, 66}, result)
+
+    def test_import_libraries(self):
+        self.rfhub_importer = RfhubImporter((self.fixture_path / 'SingleClassLib',
+                                             self.fixture_path / 'LibWithInit'), True, self.client)
+        with responses.RequestsMock() as rsps:
+            rsps.add(responses.GET, f'{self.client.api_url}/collections/', json={},
+                     status=200, adding_headers={"Content-Type": "application/json"})
+            rsps.add(responses.POST, f'{self.client.api_url}/collections/',
```
pbylicki: If you want you could make some helper method for adding this mocked json responses to make the code more concise. But it can be also done later, as different PR.

pbylicki:
@MaciejWiczk Unit testing cli could be done if you would mock RfHubImporter instance. But I think that with plans for e2e tests there is no real gain in doing it now, especially since there is not much of our own logic there.

rfhub2/cli/api_client.py
MaciejWiczk: Ok, I will get rid of it. I could use healthcheck endpoint, but I still feel that authorisation is crucial in this one. Is there a sense in having authorize endpoint? Then logic would be to have authorise in main and then to go on with repopulating app.

rfhub2/cli/rfhub_importer.py
MaciejWiczk: with method `check_commnication_with_app` that wouldn't happen. Without this method I suppose I should add try/except with break in except clause.

rfhub2/cli/api_client.py
MaciejWiczk: It didn't highlight it, probably due to import of those obcjects, but of course I will switch to Dict

rfhub2/cli/rfhub_importer.py
MaciejWiczk: Fixed

rfhub2/cli/rfhub_importer.py
MaciejWiczk: Fixed

rfhub2/cli/rfhub_importer.py
MaciejWiczk: Yeah, no problem. Fixed

rfhub2/cli/rfhub_importer.py
MaciejWiczk: Unfortunately, it was failing with Path object. 
`Building library 'F:\Docs\Desktop\rfhub-new\tests\acceptance\fixtures\SingleClassLib\SingleClassLib.py' failed: AttributeError: 'WindowsPath' object has no attribute 'split'` while it has no problems with passing path as a string.

rfhub2/cli/api_client.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        Gets list of collections object using request get method.
+        """
+        return self._get_request(endpoint='collections')
+
+    def add_collection(self, data: Collection) -> Collection:
```
MaciejWiczk: should data be hinted as Dict instead of collection? Same with other methods

rfhub2/cli/cli.py
MaciejWiczk: I misunderstood documentation. Changed to paths

rfhub2/cli/rfhub_importer.py
MaciejWiczk: Fixed

rfhub2/cli/rfhub_importer.py
MaciejWiczk: Fixed, but still:
shame on me

rfhub2/cli/rfhub_importer.py
MaciejWiczk: Yes, I know about it, and agree to keep it for later :)

tests/cli/rfhub_importer.py
MaciejWiczk: That's exactly what I thought. Should helper method return context manager?

rfhub2/cli/rfhub_importer.py
```diff
@@ -87,8 +87,9 @@ def create_collections(self, paths: Set[Path]) -> List[Dict]:
             try:
                 collection = self.create_collection(path)
                 collections.append(collection)
-            except DataError:
+            except (DataError, SystemExit) as ex:
```
MaciejWiczk, line 89: Ugly but needed. I did a check on libraries at workplace, and a lot of scripts caused issue. It's neither a fault of app nor LibDocPckg, but it's good to be prepared for such scenario.

rfhub2/cli/api_client.py
pbylicki: With Basic auth there isn't really a point, but if we would introduce different auth scheme (like LDAP), then the step for logging in should be added and then basic auth could also implement this for the sake of having unified logic. Actually adding this endpoint would be a good way of faiiling fast when user does not pass credentials and would not require handling further in the code. Let's add it, but as a next PR, leave it like this for now, ok?

rfhub2/cli/rfhub_importer.py
```diff
@@ -106,18 +104,21 @@ def add_collections(self, collections: List[Collection]) -> List[Dict[str, int]]
         loaded_collections = []
         for collection in collections:
             coll_req = self.client.add_collection(collection)
+            if 'unauthorized' in coll_req.get('detail', '').lower():
+                print(coll_req['detail'])
+                exit(1)
```
pbylicki: Maybe logging and raising StopIteration error would be cleaner solution than closing the process. Also, maybe checking for status code would be easier?

rfhub2/cli/rfhub_importer.py
pbylicki: Can you present the scenario showing why it needs to be handled by SystemExit?

pbylicki: Also, could you add unit test(s) that would cover this scenario? (DataError and ideally also something like you observed on real life execution)

MaciejWiczk: > Can you present the scenario showing why it needs to be handled by SystemExit?

Yes, the situation occured when RfHub was trying to import .py script that was using argparse and was not a file that could be translated into LibDoc.
```
import argparse

parser = argparse.ArgumentParser(prog='script_for_system_exit', description='I should trigger SystemExit')
parser.add_argument('-c', '--conf', required=True)
args = parser.parse_args()
```

> Also, could you add unit test(s) that would cover this scenario? (DataError and ideally also something like you observed on real life execution)

I will prepare scripts and unittests for such scenarios.
DataError can be triggered easily, with trying to open non-existsing file:
```
with open('non-existing-file.txt') as f:
    content = f.readlines()
```

rfhub2/cli/rfhub_importer.py
MaciejWiczk: Yeah, status code should be the way, but let me ask You before:
Should I change methods with add/delete collection/keyword so that it resturns tuple of json and return code, or should I append return code to the json itself? First one seems like proper way of doing it, and the latter looks like workaround,, to adapt to current solution.
Having in mind that this will be possibly deleted and moved to some healthcheck and authorizecheck:
>With Basic auth there isn't really a point, but if we would introduce different auth scheme (like LDAP), then the step for logging in should be added and then basic auth could also implement this for the sake of having unified logic. Actually adding this endpoint would be a good way of faiiling fast when user does not pass credentials and would not require handling further in the code. Let's add it, but as a next PR, leave it like this for now, ok?

What should I do?

rfhub2/cli/rfhub_importer.py
MaciejWiczk: After some thinking, this is not necessary complication. I removed it.

rfhub2/cli/api_client.py
```diff
@@ -54,7 +54,7 @@ def _post_request(self, endpoint: str, data: Dict) -> post:
         Sends post request to collections or keywords endpoint.
         """
         request = self.session.post(url=f'{self.api_url}/{endpoint}/', json=data)
-        return request.json()
+        return request.status_code, request.json()
```
MaciejWiczk, line 57: Doing it that way is definitely cleaner, should I add it to _get and _delete methods as well to keep consistency?

rfhub2/cli/rfhub_importer.py
MaciejWiczk: I've added returning of return code along with json in `_post_request` method. Should I add that to `_get_request` and `_delete_request` as well?

tests/cli/rfhub_importer.py
```diff
@@ -141,13 +151,13 @@ def test_add_collections_should_return_loaded_collections_and_keywords_number(se
...
     def test_add_collections_should_exit_when_unauthorized(self):
-        with self.assertRaises(SystemExit) as cm:
+        with self.assertRaises(StopIteration) as cm:
             with responses.RequestsMock() as rsps:
                 rsps.add(responses.POST, f'{self.client.api_url}/collections/',
                          json={'detail': 'Unauthorized to perform this action'},
                          status=401, adding_headers={"Content-Type": "application/json", "accept": "application/json"})
                 self.rfhub_importer.add_collections([EXPECTED_COLLECTION2])
-        self.assertEqual(cm.exception.code, 1)
+        self.assertRaises(StopIteration)
```
pbylicki: this line looks obsolete

tests/cli/rfhub_importer.py
```diff
@@ -130,6 +132,14 @@ def test_create_collection_should_return_collection(self):
         result = self.rfhub_importer.create_collection(FIXTURE_PATH / 'SingleClassLib' / 'SingleClassLib.py')
         self.assertDictEqual(EXPECTED_COLLECTION, result)
 
+    def test_create_collections_should_retunr_empty_list_on_data_error(self):
```
pbylicki: spelling errors in names of the two new test methods

rfhub2/cli/api_client.py
pbylicki: i would not expose the status code outside, and handle the auth and deserialization issues in the client. That was the initial idea for returning the pydantic model instances here and having Collection/Keyword as returned type and not just Dict. For example first you would check the status code and raise some AuthenticationError if 401 and then try to create and return pydantic model instance out of request.json() or raise DeserializationError otherwise. Then when you use the client methods you could rely on instances of Collection/Keyword object having all the required fields present and validated. And you could catch errors raised by client to reraise them or raise StopIteration, etc. without the need to add any other logic related to request and request body schema outside of the client. This should be the way how all other client methods work too and there should be shared logic between them.
This issue involves adding definition of our domain specific errors and using pydantic on cli side so if you want we can leave it like this for now and refactor it in the subsequent PR.

rfhub2/cli/rfhub_importer.py
pbylicki: As I wrote in the other comment, let's leave the status code returned for post like you did it and let's create a task for adding the approach described there. We need to wrap up this PR and start refactoring step by step.

tests/cli/rfhub_importer.py
MaciejWiczk: Removed

tests/cli/rfhub_importer.py
MaciejWiczk: Corrected

rfhub2/cli/api_client.py
MaciejWiczk: So if I understood You correctly:

1. Type annotation for this method should stay the same, but type annotation for `create_collection` and `create_keyword` should be changed to Collection and Keyword.
2. Additional logic to handle status code 401 and throw AuthenticationError.
3. Additional logic to handle Deserialization - handle status code 422 in `_post_request`.
4. Pass request.json() to `create_collection` and `create_keyword` and deserialize json to either Collection or Keyword model.
5. Change all methods to work like that.

> This issue involves adding definition of our domain specific errors and using pydantic on cli side so if you want we can leave it like this for now and refactor it in the subsequent PR.

I agree, let's have initial working version and then refactor it step by step in small PR.

rfhub2/cli/api_client.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        return self._get_request(endpoint='collections')
+
+    def add_collection(self, data: Dict) -> Dict:
+        """
+        Adds collection using request post method.
+        """
+        return self._post_request(endpoint='collections', data=data)
+
+    def delete_collection(self, id: int) -> delete:
```
pbylicki: type annotation of return value is Response from requests

rfhub2/cli/api_client.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        return self._delete_request(endpoint='collections', id=id)
+
+    def add_keyword(self, data: Dict) -> Dict:
+        """
+        Adds keyword using request post method.
+        """
+        return self._post_request(endpoint='keywords', data=data)
+
+    def _get_request(self, endpoint: str) -> get:
```
pbylicki: type annotation of return value is Dict

rfhub2/cli/api_client.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        Sends get request from given endpoint.
+        """
+        request = self.session.get(url=f'{self.api_url}/{endpoint}/')
+        return request.json()
+
+    def _post_request(self, endpoint: str, data: Dict) -> post:
```
pbylicki: type annotation of return value is Tuple[int, Dict]

rfhub2/cli/api_client.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        Sends post request to collections or keywords endpoint.
+        """
+        request = self.session.post(url=f'{self.api_url}/{endpoint}/', json=data)
+        return request.status_code, request.json()
+
+    def _delete_request(self, endpoint: str, id: int) -> delete:
```
pbylicki: type annotation of return value is Response from requests

rfhub2/cli/api_client.py
pbylicki: 1.Current type annotation should be updated - i added separate comment about it. After changes described above return value will be Collection or Keyword. method responsible for mapping dict to model instance will be called inside this method.

3.It should be both Serialization and Deserialization error - serialization is the one you described and deserialization will be when api will respond with 201, but client won't be able to create model instance from returned json.

rfhub2/cli/api_client.py
MaciejWiczk: fixed

rfhub2/cli/api_client.py
MaciejWiczk: fixed

rfhub2/cli/api_client.py
MaciejWiczk: fixed

rfhub2/cli/api_client.py
MaciejWiczk: fixed

pbylicki (approved):

PR #23, opened 2019-07-21
Adding click args and opts

MaciejWiczk:
* more click options
* remove packege dependancy on config.py
* delete option to flush DB before populating app


rfhub2/cli/population.py
```diff
@@ -1,29 +1,36 @@
...
-class LibraryPopulation(object):
+class AppPopulation(object):
 
-    def __init__(self, paths: Tuple[str, ...], no_installed_keywords: bool) -> None:
+    def __init__(self, app_interface: str, port: int, user: str, password: str,
+                 paths: Tuple[str, ...], no_installed_keywords: bool) -> None:
         self.paths = paths
-        self.auth = (BASIC_AUTH_USER, BASIC_AUTH_PASSWORD)
         self.no_installed_keywords = no_installed_keywords
+        self.client = Client(app_interface, port, user, password)
```
pbylicki: Maybe pass client instance as an argument instead of initializing it inside? It will make unit testing easier.

rfhub2/cli/client.py
```diff
@@ -0,0 +1,79 @@
...
+                  f'Check parameters for app-interace and port.')
+            exit(1)
+        req_check_auth = self._post_request('collections', TEST_COLLECTION)
+        if req_check_auth.status_code == 201:
+            self._delete_request('collections', req_check_auth.json()["id"])
+        else:
+            print(f'{req_check_auth.reason}! Check used credentials.')
+            exit(1)
+
+    def get_collections(self) -> get:
```
pbylicki: In general I would see these methods taking instances of Keyword/Collection where applicable (exact class matching those used in endpoint deifinition) and returning appropriate Keyword/Collection or list of them. If there is some error, they should raise the custom errors (should be defined in separate module) indicating to user what went wrong - connection error, not authenticated, api error, etc. with stacktrace

rfhub2/cli/client.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        Sends get request from given endpoint.
+        """
+        return self.session.get(url=f'{self.api_url}/{endpoint}', headers={"accept": "application/json"})
+
+    def _post_request(self, endpoint: str, data: Dict) -> post:
+        """
+        Sends post request to collections or keywords endpoint.
+        """
+        return self.session.post(url=f'{self.api_url}/{endpoint}/', auth=self.auth, json=data,
```
pbylicki: auth and headers can be added directly to the session object, do not need to be passed explicitly with each method if they are the same

rfhub2/cli/client.py
```diff
@@ -0,0 +1,79 @@
...
+        self.app_url = f'{PROTOCOL}{self.app_interface}:{self.port}'
+        self.api_url = f'{self.app_url}/{API_V1}'
+
+    def check_communication_with_app(self) -> None:
+        try:
+            self.get_collections()
+        except exceptions.RequestException:
+            print(f'Connection to application at {self.app_url} refused!\n'
+                  f'Check parameters for app-interace and port.')
+            exit(1)
```
pbylicki: Again, I would prefer raising error with meaningful message without manual exiting. It will be easier to test.

rfhub2/cli/client.py
pbylicki: If your application will not handle the error or reraise different, custom error with meaningful message it will result with application exiting with code 1 just like it does when you explicitly call `exit()`. The advantage of just raising the error is easy testing of the method raising error and not explicitly exiting the whole process. If we would want to eventually avoid displaying any stacktrace to user, i would put handling any errors with some output and exit code 1 in one place close to program main method to not have this exits scattered in various places in the code.

pbylicki (approved):
Can be merged to continue work on single branch

rfhub2/cli/cli.py
```diff
@@ -1,17 +1,35 @@
...
+from rfhub2.cli.rfhub_importer import RfhubImporter
 
 
 @click.command()
-@click.argument('paths', nargs=-1, type=click.Path(exists=True))
-@click.option('--no-installed-keywords', 'no_installed_keywords', default=False,
-              help='do not load some common installed keyword libraries, such as BuiltIn.')
-def main(paths: Tuple[str, ...], no_installed_keywords: bool) -> None:
-    """Package to populate rfhub2 with libraries and resource documentation"""
+@click.option('-i', '--app-interface', type=click.STRING, default="http://0.0.0.0",
```
pbylicki, line 9: I think that for client we should join interface and port in one option like url. This division had more sense when cli was also used for running api itself. Now it seems pointless. Also, 0.0.0.0 should be changed to localhost, it is more sensible from client perspective

PR #24, opened 2019-07-30
Add Healthcheck endpoint and fix UI bug

pbylicki:
Added healthcheck endpoint that verifies that app is running and it can connect to db. In the case of db failure endpoint will respond with 503 and failure response.
As a bonus I fixed UI so if you open the link with search pattern present in the UI, the search results will be presented. The code looks like it was intended before but not checked properly.


MaciejWiczk (approved):
Nothing to complain about :)

Issue #21, opened 2019-07-20
Rfhub starts with all collections expanded


MaciejWiczk:
Rfhub starts with all collections expanded, which might be confusing for user, and it's making it bit harder to navigate.
Current state on the left, desired state on the right:
![rfhub-new_issue](https://user-images.githubusercontent.com/35737166/61575068-8ea67c00-aac7-11e9-9e26-d63ddf0b34c7.jpg)

PR #25, opened 2019-08-03
Adding style display: none;

MaciejWiczk:
PR for issue #21


rfhub2/templates/twocolumn.html
```diff
@@ -7,7 +7,7 @@
             <li>
               <label class="tree-toggler nav-header"
                      title="file path: {{collection.path}}">{{collection.name}}</label>
-              <ul class="list-group tree">
+              <ul class="list-group tree" style="display: none;">
```
pbylicki: we do not have styles defined inline in html templates. I would put it either to css file or to `doc.js` as a default state, next to the logic responsible for toggling.

MaciejWiczk: I'm sure You're right, but I'm clueless regarding frontend. Would such a form be alright:
```
.list-group.tree {
    display: none;
}
```
added to bootstrap.min.css file

pbylicki: it is correct, but should be added to `doc.css` which has all the custom style definitions, bootstrap.min.css is third party and we should not add anything to it.

PR #35, opened 2019-08-29
new frontend poc

pbylicki:
Current version of new frontend app is available here with example data loaded: https://rfhub2-demo.herokuapp.com
Of course there is some polishing to be done, but I think that it provides all functionalities of the old frontend and can already substitute it.


.travis.yml
```diff
@@ -13,3 +13,4 @@ deploy:
   provider: heroku
   app: rfhub2-demo
   api_key: $HEROKU_AUTH_TOKEN
+  on: frontend-poc
```
pbylicki: this line is only for automatic demo purpose and should be removed before merge

MaciejWiczk:
This looks really awesome! It's mind blowing. I see just one tiny thing from old UI that could be useful - if top left rfhub2 could redirect to main page.
I will play around with it and deploy at work on openshift :)

MaciejWiczk (approved):
Awesome!

MaciejWiczk:
I've played with it at work and it's awesome. I can think of two enhancement for future - left panel could be wider, and app could display library version in collection overview as well as on homepage.

pbylicki:
@MaciejWiczk 
- displaying version added
- drawer is 25% wider
- app title is a link to main page
- Readme extended with current building instructions

PR #36, opened 2019-08-31
Added library scope to display of collection overview

MaciejWiczk:
Added library scope to display of collection overview. I think it would be good to have this fucntionality as well as version displayed. @pbylicki could You take a look at [js1](https://github.com/pbylicki/rfhub-new/compare/feature/add_scope_to_display?expand=1#diff-fe74fa2246e3001c8304953fda6797f8) and [js2](https://github.com/pbylicki/rfhub-new/compare/feature/add_scope_to_display?expand=1#diff-a2a7100bd561f90064d8f3085a6ffdd4)
git is not handling diff very well of such files. I did add:
```
,y.a.createElement("div",null,"scope: ",t.detailCollection.scope)
``` 
and:
```
type: string\n  scope: string | null
```
and:
```
\n          <div>scope: {store.detailCollection.scope}</div>
```
I don't know if it's the proper way, but somehow it's working.


pbylicki:
@MaciejWiczk this js files generated as build wouldn't be normally stored in version control. Once I will manage to add building frontend to pipeline all these static files will probably be removed.
Did you generated them using the script I added in last PR?

MaciejWiczk:
> Did you generated them using the script I added in last PR?

Nope, just manual labour ;)
I've seen it but forgot. I will check what would be result of script compared to manual changes.

pbylicki:
> Nope, just manual labour ;)

Respect, but please use the script `build_ui.sh`. It should remove old files and put new ones in their place. then they can be commited

MaciejWiczk:
> Respect

Just a Franklin journey to understand js ;)

> please use the script

Done

pbylicki (approved):

PR #41, opened 2019-09-05
Add simple implementation of infinite list

### LINKED ISSUE #29: RfHub2 is not displaying more than 100 items in left panel
RfHub2 is not displaying more than 190 items in left panel. 
HTR:
run application and populate it
```
for i in {1..500}; do curl -u rfhub:rfhub -X POST "http://localhost:8000/api/v1/collections/" -H "accept: application/json" -H "Content-Type: application/json" -d "{\"name\":\"$i\",\"type\":\"string\",\"version\":\"string\",\"scope\":\"string\",\"named_args\":\"string\",\"path\":\"string\",\"doc\":\"string\",\"doc_format\":\"string\"}"; done
```

Actual state:
rfhub_bug

pbylicki:
Fixes #29 for all list views in the application
I tried to use some out-of-the-box solutions, but there seemed to be some issues due to mobx usage (I think). So I settled down with simple loading data when user scrolls through last few records on the list.


MaciejWiczk (approved):
Awesome. Already tested on 160 collections with no problems :)

Issue #48, opened 2019-09-08
Add setup.py for rfhub2


MaciejWiczk:
Add setup.py for rfhub2, with some scripts to build and upload package to PyPi.

PR #49, opened 2019-09-08
Changes for packaging

MaciejWiczk:
Fix for #48. Scripts comes mostly from rfhub2-old. I changed setup.py and MANIFEST mostly. 
Added:
* setup.py, 
* LICENSE, 
* MANIFEST.in,
* publish.pypi

Changed __main__.py to call main() function.
Bumped version to one above that on pypi.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,12 +16,9 @@ setup(
     long_description_content_type='text/markdown',
     zip_safe=True,
     include_package_data=True,
-    install_requires=['aiofiles==0.4.0', 'Click==7.0', 'dataclasses==0.6', 'fastapi==0.30.0', 'h11==0.8.1',
-                      'httptools==0.0.13', 'jinja2==2.10.1', 'MarkupSafe==1.1.1', 'pydantic==0.28',
-                      'robotframework==3.1.2', 'SQLAlchemy==1.3.3', 'starlette==0.12.0', 'uvicorn==0.7.1',
-                      'uvloop==0.12.2', 'websockets==7.0', 'requests==2.22.0'],
+    install_requires=['Click>=7.0', 'fastapi>=0.30.0', 'robotframework>=3.0.0', 'SQLAlchemy>=1.1.0', 'requests>=2.10.0'],
     extras_require={
-        "postgresql": ["psycopg2-binary==2.8.3"]
+        "postgresql": ["psycopg2-binary>=2.7.4"]
     },
     classifiers=[
         "Development Status :: 4 - Beta",
@@ -55,7 +52,7 @@ setup(
     entry_points={
         'console_scripts': [
             "rfhub2 = rfhub2.__main__:main",
-            "rfhub2.cli = rfhub2.cli.__main__:main",
+            "rfhub2-cli = rfhub2.cli.__main__:main",
         ]
     }
 )
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,61 @@
+from setuptools import setup
+
+filename = 'rfhub2/version.py'
+exec(open(filename).read())
+
+setup(
+    name='rfhub2',
+    version=version,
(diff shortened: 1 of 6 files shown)


MaciejWiczk:
After I fix #3, CICD could be changed, to automatically upload package to PyPi and docker images to docker repository.

setup.py
```diff
@@ -0,0 +1,61 @@
...
+    author_email=['pawelkbylicki@gmail.com', 'maciejwiczk@gmail.com'],
+    url='https://github.com/pbylicki/rfhub2/',
+    keywords='robotframework',
+    license='Apache License 2.0',
+    description='Webserver for robot framework and python assets documentation',
+    long_description=open('README.md').read(),
+    long_description_content_type='text/markdown',
+    zip_safe=True,
+    include_package_data=True,
+    install_requires=['aiofiles==0.4.0', 'Click==7.0', 'dataclasses==0.6', 'fastapi==0.30.0', 'h11==0.8.1',
```
pbylicki: I would reduce it to only top-level dependencies, I think it is too strict and does not allow for Windows installation. It could be reduced to Click, fastapi, robotframework, sqlalchemy and requests. I would also relax the requirements for versions, and if required put lower-than/greater-than kind of constraints. It would require some testing, I think that there may be some breaking changes for fastapi dependencies since the development is heavy there. for robot probably version 3+ is enough. For the rest I think that any recent version should work.

setup.py
```diff
@@ -0,0 +1,61 @@
...
+                      'httptools==0.0.13', 'jinja2==2.10.1', 'MarkupSafe==1.1.1', 'pydantic==0.28',
+                      'robotframework==3.1.2', 'SQLAlchemy==1.3.3', 'starlette==0.12.0', 'uvicorn==0.7.1',
+                      'uvloop==0.12.2', 'websockets==7.0', 'requests==2.22.0'],
+    extras_require={
+        "postgresql": ["psycopg2-binary==2.8.3"]
```
pbylicki: Same comment about strict version as above

setup.py
MaciejWiczk: Sure thing. I will change it to more relaxed way. We can introduce dependabot or PyUp for managing dependencies. Probably after #3. What do You think?

setup.py
```diff
@@ -0,0 +1,61 @@
...
+        'rfhub2.db.repository',
+        'rfhub2.model',
+        'rfhub2.ui',
+        'rfhub2.utils',
+    ],
+    scripts=[],
+    entry_points={
+        'console_scripts': [
+            "rfhub2 = rfhub2.__main__:main",
+            "rfhub2.cli = rfhub2.cli.__main__:main",
```
pbylicki: I was wondering if we can make this executable name more compact and still recognizable, but i couldn't come up with anything. I was just thinking that maybe in could be switched to `rfhub2-cli` to not mistake cli for file extension

pbylicki:
> After I fix #3, CICD could be changed, to automatically upload package to PyPi and docker images to docker repository.

I was thinking on creating the script to publish the package as step in master pipeline only if the version specified in rfhub2.version is higher than existing latest version on pypi. This way it would be automatic, but we would be able to control when new version is published, not necessarily on each change to master. What do you think?

setup.py
MaciejWiczk: Sure thing!

MaciejWiczk:
> > After I fix #3, CICD could be changed, to automatically upload package to PyPi and docker images to docker repository.
> 
> I was thinking on creating the script to publish the package as step in master pipeline only if the version specified in rfhub2.version is higher than existing latest version on pypi. This way it would be automatic, but we would be able to control when new version is published, not necessarily on each change to master. What do you think?

I had very same thing on mind. For now let's upload it manually, when there is a need.

pbylicki (approved):

PR #56, opened 2019-09-17
Added aiofiles and uvicorn to setup.py, tested locally, and it's working

MaciejWiczk:


pbylicki:
please don't tell me that previous builds did not work because there was no uvicorn :D

pbylicki (approved):

MaciejWiczk:
> please don't tell me that previous builds did not work because there was no uvicorn :D

yeah...I will speed up with robot - acceptance tests. How about usin some bare docker image and installing package inside?

pbylicki:
@MaciejWiczk  yes, acceptance must be run on installed package. Maybe to not require docker in CI pipeline after package is built it could be installed and run inside virtualenv? then it could be deleted after tests are executed.

PR #58, opened 2019-10-02
Install rfhub2 in venv for acceptance test

pbylicki:


tests/acceptance/resources/keywords.resource
```diff
@@ -1,88 +1,90 @@
...
-
+*** Settings ***
+Library    SeleniumLibrary
+Library    OperatingSystem
+Library    Process
+Library    RequestsLibrary
+Library    Collections
+
+*** Variables ***
+${activate_venv}    . atests-venv/bin/activate &&
```
pbylicki, line 9: This should be more customizable to allow running without the separate venv. Do you have an idea how to manage this? Maybe something like what we had for switching between sqlite/postgres db for tests in old repository?

MaciejWiczk: If You mean [this](https://github.com/pbylicki/rfhub2-old/blob/master/tests/conf/default.args) than yeas, this is smooth. If I understand this correctly, this is made to run tests for SQLite and PostgreSQL in one venv, but separate from one used for building and unittests?
PS. Should this be overriden when passing config file?

MaciejWiczk (approved):

tests/acceptance/resources/keywords.resource
pbylicki: @MaciejWiczk Yes, it is making process running rfhub2 webserver and rfhub2 cli in different virtualenv than the rest of the CI pipeline and effectively checking if pip installation works properly.
I think it could be handled similarly as DB_URI in one of last PRs. So by default it would be defined as empty, and overriden by config file value. It would allow for running the tested application in current env/virtualenv in development if required and in isolated virtualenv in CI.
Would you like to implement it?
PS. This setup would also allow testing different installation modes in separate virtualenvs (SQLite-only, postgres), but I do not think that it is worth the effort for now. But still it is something worth to remember for the future or other projects.

MaciejWiczk: I will take it on. I'm thinking that later on it might be changed into env variable if we would like to use build matrix with python 3.7 and 3.8

MaciejWiczk (approved):

PR #66, opened 2019-10-19
Feature/rfhub 64 add update option

### LINKED ISSUE #64: Rfhub2 could keep all collections until load of rfhub2-cli is finished
Currently rfhub2-cli deletes all collections with keywords ant then starts with app population. In environment that has a lot of collections and keywords, this results in service temporarly showing only some of them. 

- This could be aided by creating temporary table for collections/keywords, and switching them when rfhub2-cli is done. This would require to add (private?) endpoint triggering tables switch.

- Alternatively app could delete keywords and collections from original tables, that are not in temporary ones (probably by checking names and paths of collections, or versions, then comparing keywords inside collections) and then doing upsert with SQL merge.

- Last option is to use update endpoint, for both collections and keywords

First way seems easier than second, and third seems like cleanes solution. @pbylicki what do You think?

MaciejWiczk:
Fixes #64. Changes done in order to allow for update mode, that will update existing libraries, delete obsolete and add new ones. 
rfhub-cli package main change is mode introduction, instead of `--no-db-flush`. 
Mode can be one of 
- `insert` - recreates all collections in app-, default one, 
- `append` - adds every collection to app without any removal
- `update` - updates existing libraries, deletes obsolete and adds new ones. 

I had a go with using update endpoint, but the logic I came up with was no good. Three way compare of collections and collections keywords wasa bit too much in one go. I will think of refactoring this implementation into new one.


rfhub2/cli/rfhub_importer.py
```diff
@@ -264,3 +317,124 @@ def _has_test_case_table(data: str) -> bool:
...
+        }
+
+    @staticmethod
+    def _get_outdated_collections(
+        existing_collections: List[Dict], new_collections: List[Dict]
+    ) -> Set[int]:
+        """Returns set of collection ids that were found in application but are outdated"""
+        outdated_collections = set()
+        if len(existing_collections) > 0:
+            outdated_collections = set()
```
MaciejWiczk: delete this line

rfhub2/cli/rfhub_importer.py
```diff
@@ -264,3 +317,124 @@ def _has_test_case_table(data: str) -> bool:
...
+        )
+
+    @staticmethod
+    def _library_or_resource_changed(
+        new_collection: Dict, existing_collection: Dict
+    ) -> bool:
+        if new_collection["type"] == "library":
+            return new_collection["version"] != existing_collection["version"]
+        else:
+            return new_collection != existing_collection
```
MaciejWiczk: Way too sensitive. Needs to be changed to first compare resource overview and keywords as list for changes, not as a single item. Ie two `all()`  for twa way compare

rfhub2/cli/rfhub_importer.py
```diff
@@ -39,6 +44,16 @@ def delete_all_collections(self) -> Set[int]:
             collections_id.update(self._delete_collections())
         return collections_id
 
+    def get_all_collections(self) -> List[Dict]:
+        """Gets all collections from application"""
+        collections = []
+        for i in range(0, 999999, 100):
+            collection_slice = self.client.get_collections(i, i + 100)
```
MaciejWiczk: remove `i + `

rfhub2/cli/rfhub_importer.py
```diff
@@ -39,6 +44,16 @@ def delete_all_collections(self) -> Set[int]:
             collections_id.update(self._delete_collections())
         return collections_id
 
+    def get_all_collections(self) -> List[Dict]:
+        """Gets all collections from application"""
+        collections = []
+        for i in range(0, 999999, 100):
```
pbylicki, line 50: If we rely on the collection slice size to determine if there are any results left, we could do this without this limit and just call the method and append as long as the next requests are having any results, you can do this only based on page size.
Also, 100 seems reasonable given that some of collections can have very large output, maybe you can later test if it wouldn't be faster to have bigger page when working in a corporate network :)

rfhub2/cli/rfhub_importer.py
```diff
@@ -56,7 +71,17 @@ def import_libraries(self) -> Tuple[int, int]:
...
         collections = self.create_collections(libraries_paths)
-        loaded_collections = self.add_collections(collections)
+        if self.mode == "append":
+            loaded_collections = self.add_collections(collections)
+        elif self.mode == "insert":
+            self.delete_all_collections()
+            loaded_collections = self.add_collections(collections)
+        else:
+            existing_collections = self.get_all_collections()
+            self.delete_outdated_collections(existing_collections, collections)
```
pbylicki: Maybe delete step could be put after the update step? If there were some errors with update, we would still have data from previous executions. I think that if I were to choose, having partial update without old data loss is better. I think that in the future it would be best to move the update logic to the server endpoint to make this operation safer for data integrity.

rfhub2/cli/rfhub_importer.py
```diff
@@ -264,3 +317,134 @@ def _has_test_case_table(data: str) -> bool:
...
+
+    @staticmethod
+    def _get_obsolete_collections(
+        existing_collections: List[Dict], new_collections: List[Dict]
+    ) -> Set[int]:
+        """Returns set of collection ids that were found in application but not in paths"""
+        return {
+            existing_collection["id"]
+            for existing_collection in existing_collections
+            if all(
```
pbylicki: You could create set of new_collection paths and then check if it contains existing collection path

rfhub2/cli/rfhub_importer.py
```diff
@@ -264,3 +317,134 @@ def _has_test_case_table(data: str) -> bool:
             re.search(r"^\*+\s*(Test Cases?)", data, re.MULTILINE | re.IGNORECASE)
             is not None
         )
+
+    @staticmethod
+    def _get_obsolete_collections(
```
pbylicki: maybe the methods that only return ids could have method names adjusted to reflect it

rfhub2/cli/rfhub_importer.py
```diff
@@ -264,3 +317,134 @@ def _has_test_case_table(data: str) -> bool:
...
+
+    @staticmethod
+    def _get_new_collections(
+        existing_collections: List[Dict], new_collections: List[Dict]
+    ) -> List[Dict]:
+        """Returns list of collections to insert that were found in paths but not in application"""
+        return [
+            new_collection
+            for new_collection in new_collections
+            if all(
```
pbylicki: can be refactor to set of existing collection paths

rfhub2/cli/rfhub_importer.py
```diff
@@ -264,3 +317,134 @@ def _has_test_case_table(data: str) -> bool:
...
+            new_collection["keywords"], reduced_collection["keywords"]
+        )
+        return reduced_collection
+
+    @staticmethod
+    def _get_reduced_collection(
+        new_collection: Dict, existing_collection: Dict
+    ) -> Dict:
+        """Returns existing_collection dictionary with key/value pairs reduced to the ones from new_collection"""
+        return {k: existing_collection[k] for k in new_collection.keys()}
```
pbylicki: can we have a case when new collection has key not present in existing collection?

rfhub2/cli/rfhub_importer.py
```diff
@@ -264,3 +317,134 @@ def _has_test_case_table(data: str) -> bool:
...
+
+    @staticmethod
+    def _get_reduced_keywords(
+        new_collection_keywords: List[Dict], existing_collection_keywords: List[Dict]
+    ) -> List[Dict]:
+        if min(len(new_collection_keywords), len(existing_collection_keywords)) > 0:
```
pbylicki, line 414: the opposite case should be handled and return a list too.

rfhub2/cli/rfhub_importer.py
```diff
@@ -264,3 +317,134 @@ def _has_test_case_table(data: str) -> bool:
...
+
+    @staticmethod
+    def _library_or_resource_changed(
+        new_collection: Dict, existing_collection: Dict
+    ) -> bool:
+        if new_collection["type"] == "library":
+            return new_collection["version"] != existing_collection["version"]
+        else:
+            return (
+                any(
```
pbylicki, line 442: first two conditions could be written as set's symmetric difference

rfhub2/cli/rfhub_importer.py
```diff
@@ -264,3 +317,134 @@ def _has_test_case_table(data: str) -> bool:
...
+                    keyword not in new_collection["keywords"]
+                    for keyword in existing_collection["keywords"]
+                )
+                or any(
+                    keyword not in existing_collection["keywords"]
+                    for keyword in new_collection["keywords"]
+                )
+                or {k: v for k, v in new_collection.items() if k != "keywords"}
```
pbylicki: could you explain why these two conditions are needed?

rfhub2/cli/rfhub_importer.py
MaciejWiczk: > Also, 100 seems reasonable given that some of collections can have very large output, maybe you can later test if it wouldn't be faster to have bigger page when working in a corporate network :)

Locally it's a deamon of speed, but on my corporate network it is serving 100 collections in 1.5 second on OC with postgreSQL :( It's a matter of network speed/database response rather than app itself. App is granted with 2 CPUs and 1 GB of RAM, and it's consumig 0.4 CPU and 75 MB of memory at max.

rfhub2/cli/rfhub_importer.py
MaciejWiczk: > Maybe delete step could be put after the update step?

Good point. Better to have collections twice than not at all.



> I think that in the future it would be best to move the update logic to the server endpoint to make this operation safer for data integrity.

You mean like to have temporary tables and than swap or something totally different?

rfhub2/cli/rfhub_importer.py
MaciejWiczk: I'm rather confident it won't happen, but we can use get method with some default, like '' or None.

rfhub2/cli/rfhub_importer.py
MaciejWiczk: Valid point, I somehow missed that

rfhub2/cli/rfhub_importer.py
MaciejWiczk: Sets would be smooth and fast, however `*_collection["keywords"]` are lists of dicts, thus cannot be simply converted to sets.

rfhub2/cli/rfhub_importer.py
MaciejWiczk: Yes, those two conditions are to check if overall collction documentation hasn't changed. Removing keywords from collection keys is needed when keywords are the same, but in different order, thus giving false positive feedback, that collections has changed. Without keywords list, comparing `existing_collection` and `new_collection` can be handled simply like that.

I think that this method should be moved to some helper method and named properly.

rfhub2/cli/rfhub_importer.py
MaciejWiczk: Clever, no need to iterate n*m

rfhub2/cli/rfhub_importer.py
MaciejWiczk: I removed limit from this method

rfhub2/cli/rfhub_importer.py
MaciejWiczk: I swapped order of commands

rfhub2/cli/rfhub_importer.py
MaciejWiczk: done

rfhub2/cli/rfhub_importer.py
MaciejWiczk: I agree, fixed names where needed

rfhub2/cli/rfhub_importer.py
MaciejWiczk: done

rfhub2/cli/rfhub_importer.py
MaciejWiczk: Extracted part of it to separate method

pbylicki (approved):

PR #119, opened 2019-12-21
Change Statistics model and basic endpoints

### LINKED ISSUE #67: Extending Rfhub2 with analytics
Currently Rfhub is doing it's job being a keyword interactive search app. What seems like natural continuation is providing users with some analytics data from tests executions:

- how many times each library/keywords were used. This could help to identify non-used collections, that could be removed, and heavy used collections, that could be converted into library, refactored etc.
- what are average times of keyword execution, so it may be refactored and improved if it's slow.

Implementing such changes would require:

Changes on backaed:
- new package, something like DBBot, but just for keyword statistics, 

Changes on frontend:
- first page - more columns in table, new column would show how many times library was used (sum of keywords usage from this library), maybe with some visual indicator, like progressbar.
- changes on library details page, with number of keyword execution times, visual indicator and average keyword execution time.
- changes to tables, so they can be sorted by each column.

@pbylicki, what do You think about that?

pbylicki:


MaciejWiczk:
@pbylicki what do we do about this one? Should we merge it to master to avoid conflicts later on, or will You work on it, and merge should wait?

pbylicki:
@MaciejWiczk I wouldn't merge it yet since there still may be some changes in the API and we will be releasing some smaller changes in the meantime. In particular I would like to create and test the endpoints that would return collections and keywords with statistics included (aggregated per request, not cached yet) to see if it can be used as initial implementation and base for UI changes.

rfhub2/api/endpoints/keywords.py
```diff
@@ -32,9 +39,14 @@ def get_keywords_with_stats(
     limit: int = 100,
     pattern: str = None,
     use_doc: bool = True,
+    collection_id: Optional[int] = None,
 ):
     return repository.get_all_with_stats(
-        skip=skip, limit=limit, pattern=pattern, use_doc=use_doc
+        skip=skip,
```
MaciejWiczk, line 50: Sorry for not noticing it earlier. Due to different endpoint app is showing only first 100 keywords. Maybe when collection_id is not None we could skip the limit, to have same experience as previosly?

pbylicki: good point, I will set UI to use large limit for now

MaciejWiczk:
@pbylicki what is preventing from merging this one? Do You have some other plans for that one?

pbylicki:
@MaciejWiczk did you have a chance to test it with some real life data to see how adding the statistics calculation affects the performance?
Apart from merging all the latest changes I would rename Statistics to KeywordStatistics across the app to be more specific and prepare for different kinds of statistics in the future

MaciejWiczk:
> @MaciejWiczk did you have a chance to test it with some real life data to see how adding the statistics calculation affects the performance?

I'm on holidays till next friday, so it will have to wait till then. 

> Apart from merging all the latest changes I would rename Statistics to KeywordStatistics across the app to be more specific and prepare for different kinds of statistics in the future.

Sounds very reasonable. I could create PR with those changes and I will resolve conflicts on the way.

pbylicki:
@MaciejWiczk Renaming resolved in #166

MaciejWiczk:
@pbylicki I did some tests on "prod" env in company and I've noticed one thing: Posts requests are getting slower and slower with every row there is added into table in DB. I believe it is happening due to recalculation of indices. One option for mitigation would be to add an endpoint, that would accept more than one statistic object, or we could block indices recalculation on that table. I think that first approach would be nicer, but I will investigate indices recalculation as well.

pbylicki:
@MaciejWiczk Thanks for testing, did you see any issues with response time of endpoints reading collections and keywords with aggregated statistics?
I initially wanted to create endpoint for adding many statistics record, but it was lost at some point. Thanks for reminding me. I will add this before merging this PR. It should partially address the slowness. You are right that index on this table is to blame.

MaciejWiczk:
> did you see any issues with response time of endpoints reading collections and keywords with aggregated statistics?

frankly speaking I didn't think about it, but I will give it a go when first occasion occurs. 

> You are right that index on this table is to blame.

Manually adding all the record by SQL IDE, either in one statement or multiple single row inserts didn't cause this behaviour. But this is probably by IDE design tough.

MaciejWiczk:
> > did you see any issues with response time of endpoints reading collections and keywords with aggregated statistics?
> 
> frankly speaking I didn't think about it, but I will give it a go when first occasion occurs.
> 
> > You are right that index on this table is to blame.
> 
> Manually adding all the record by SQL IDE, either in one statement or multiple single row inserts didn't cause this behaviour. But this is probably by IDE design tough.

So I did tests on corporate network with fair amount of keywords. This time, having app running on localhost and Postgres DB running on OC turned out to behave linearly. 1.5 GB of outputs.xml were transformed into  ~37k rows in DB in about 3 hours. Maybe previous time I spot some net bottleneck that was fixed. To wrap up, this time it's behaving correctly, data like libraries, keywords are loading the same way they did, with the same speed. Statistics could be loaded via some endpoint that allows batch loading, as You mentioned before. I think that adding some progressbar would be beneficial for user, to indicate that app is working correctly and adding data. I will create PR with that.

pbylicki:
@MaciejWiczk I have a branch with endpoint for adding batch statistics, just need to update cli part to use it. Should be ready this week

MaciejWiczk:
> Should be ready this week.

Awesome, I can take care of cli. Can You check my PR #180 for merging it into yours? It will be simpler to work on synchronized branch.

MaciejWiczk:
@pbylicki, could You bump version before merging that one? Or do we have to do something before that?

pbylicki:
@MaciejWiczk I want to make client upload statistics in batches before the merge

MaciejWiczk:
> @MaciejWiczk I want to make client upload statistics in batches before the merge

Should I take this one #197 ?

pbylicki:
@MaciejWiczk If you have time in the next couple of days then go ahead. Otherwise I can do this at the end of this week.

MaciejWiczk:
> @MaciejWiczk If you have time in the next couple of days then go ahead. Otherwise I can do this at the end of this week.

I will give it a go today or tomorrow, I will come back with either a PR or cry for help ;)

Issue #128, opened 2020-01-03
Add clear button for search box


mnojek:
Would be nice to have some kind of clear button next to search box to quickly remove the content of it instead of removing the sentence manually.

Issue #126, opened 2020-01-03
Searching box doesn't work when [Enter] is pressed


mnojek:
If I put some value into search box and then start doing something else, the box keeps the value, but when I focus on non-empty search box again and hit [Enter] key, the searching does not execute - I need to change the value of search box to dynamically execute searching again. I propose implement running searching when the [Enter] key is pressed.

PR #135, opened 2020-01-06
Add searching on Enter and clear search button

pbylicki:
Addresses issues #126 and #128 to extend search bar functionality


MaciejWiczk:
I've run app and it's working exactly as expected. I think You could update version in version.py so we would have new release in pypi.

MaciejWiczk (approved):
I've run app and it's working exactly as expected. I think You could update version in version.py so we would have new release in pypi.

pbylicki:
@MaciejWiczk I plan to work on other frontend improvements in the upcoming days and will prepare release then

PR #139, opened 2020-01-12
Show keyword statistics on CollectionDetails view

pbylicki:


MaciejWiczk (approved):
This is awesome looking. With that, I believe we are almost done with statistics, just extendind E2E tests left.

PR #143, opened 2020-01-13
Adding conditional for returning all keywords with stats when c

MaciejWiczk:
Adding conditional for returning all keywords with stats when collection_id is not None


pbylicki (changes requested):
I don't think that this is the right approach. Such change in the endpoint behavior based on selected filter is not intuitive and is creating a limitation that should be handled on the client side like it is done for other endpoints. I would propose to currently adjust UI to call this enpoint with limit=10000. Later we would either handle paging for keywords on this view or if we will implement saving cached statistics in keywords and collection tables, we won't need additional request to get keywords.

Issue #133, opened 2020-01-03
Long names for keyword or collection display issues


mnojek:
When the collection or keyword name are too long to fit the width of the left-side menu they are not displayed properly.

![image](https://user-images.githubusercontent.com/7412964/71718869-e0945180-2e1c-11ea-813f-800a8a926c35.png)

I propose adding ellipsis (...) at the end of the too long word:

![image](https://user-images.githubusercontent.com/7412964/71719059-6a441f00-2e1d-11ea-83eb-29ddc410cbbe.png)

Or break the word in the place where it meets with the div border:

![image](https://user-images.githubusercontent.com/7412964/71719116-919aec00-2e1d-11ea-8934-d1335d02cf1f.png)

[Here is the example of solution](https://www.codelord.net/2013/08/23/css-tip-overflowing-with-text/).

MarcinMaciaszek:
elipsis (...) and ful ltext on hover would be great :)

MarcinMaciaszek:
Merged and working, close :)

Issue #130, opened 2020-01-03
Keyword button from left-side menu works only when clicked on text


mnojek:
Button representing specific keyword under the collection in left-side menu works only when it is clicked on text - when clicked outside the text but still on the button it only triggers click animation but does not open this keyword documentation.

PR #144, opened 2020-01-13
Added elipsis to long keyword names and tooltip to keywords(#133)

MarcinMaciaszek:


frontend/src/components/DrawerCollectionList.tsx
```diff
@@ -41,6 +42,18 @@ const DrawerCollectionListItem: React.FC<DrawerCollectionListItemProps> = observ
...
 
+  function hasWhiteSpace(text: string): boolean {
+    return /\s/g.test(text);
+  }
+
+  function wrapTextWithoutSpaces(text: string): string {
+    if (hasWhiteSpace(text)) {
+      return text
+      } else {
+        return `${text.substring(0, 30)}...`
```
pbylicki: it adds "..." to every keyword name, that does not have whitespace including short names like "Setup"

frontend/src/components/DrawerCollectionList.tsx
```diff
@@ -41,6 +42,18 @@ const DrawerCollectionListItem: React.FC<DrawerCollectionListItemProps> = observ
     return store.drawerSelectedCollection === collectionId
   }
 
+  function hasWhiteSpace(text: string): boolean {
+    return /\s/g.test(text);
+  }
+
+  function wrapTextWithoutSpaces(text: string): string {
+    if (hasWhiteSpace(text)) {
```
pbylicki: Maybe we could use ellipsis for any name longer than 30, even with whitespaces and to display full name on the tooltip?
Scenario with 30 letter word without spaces seems very rare, so this condition won't be really used. Maybe we could apply the same logic to collection names? @MaciejWiczk what do you think?

frontend/yarn.lock
```diff
@@ -5,14 +5,12 @@
 "@babel/code-frame@7.5.5", "@babel/code-frame@^7.0.0", "@babel/code-frame@^7.5.5":
   version "7.5.5"
   resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.5.5.tgz#bc0782f6d69f7b7d49531219699b988f669a8f9d"
-  integrity sha512-27d4lZoomVyo51VegxI20xZPuSHusqbQag/ztrBC7wegWoQ1nLREPVSKSW8byhTlzTKyNE4ifaTA6lCp7JjpFw==
```
pbylicki: Are these changes due to different yarn versions that we are using? What version do you have?

frontend/src/components/DrawerCollectionList.tsx
MaciejWiczk: You've got a point. The longest I've used is probably DatabaseLibraryUtils, which again, is a library, that could have the same logic applied.

frontend/src/components/DrawerCollectionList.tsx
MaciejWiczk: @MarcinMaciaszek, @nojas10 Could You check behaviour of those changes on real life corpo repo? ;)

frontend/yarn.lock
MarcinMaciaszek: I have yarn 1.5.0, it happened after react-tooltip installation. Maybe I'm using the newest version of yarn, so it can remove all integrity values.

frontend/src/components/DrawerCollectionList.tsx
MarcinMaciaszek: @pbylicki @MaciejWiczk 
It should work now well. As you mentioned above. Not well solved, but done :D

frontend/yarn.lock
pbylicki: yarn 1.5.0 is almost 2 years old. Could you update to latest release and rebuild?

frontend/src/stores/CollectionStore.ts
```diff
@@ -67,6 +67,15 @@ export class CollectionStore {
       })
   }
 
+  @action.bound
+  wrapTextWithoutSpaces(text: string): string {
```
pbylicki: I would keep this method in the component. It is only used there and is not directly using any values from the store.

frontend/src/components/DrawerCollectionList.tsx
pbylicki: @MaciejWiczk @MarcinMaciaszek So the idea is to have tooltip for all keywords/collections or only to those that are longer than 30 and have ellipsis? For now it is ok, but if we want the second option I could adjust it while working on #130 which affects the same part of the code

frontend/yarn.lock
MarcinMaciaszek: Updated, Have an issue with wrong yarn version used. By default was using 1.5.0 instead of 1.21.1

frontend/src/stores/CollectionStore.ts
MarcinMaciaszek: Removed, used external lib

frontend/src/components/DrawerCollectionList.tsx
MarcinMaciaszek: @pbylicki @MaciejWiczk 
Finally fixed all issues :P sorry have no time
It should work well now, so you can decide if we should merge it.

MaciejWiczk:
I pushed one commit to fix test. CI passed

MaciejWiczk (approved):
@MarcinMaciaszek, it's lookin'good! :+1:

pbylicki (approved):

frontend/src/components/DrawerCollectionList.tsx
pbylicki: Thanks, good job. Merging it right now.

PR #151, opened 2020-01-16
Add sorting support to all sortable endpoints

pbylicki:


rfhub2/db/repository/collection_repository.py
```diff
@@ -74,12 +76,13 @@ def get_all_with_stats(
...
         skip: int = 0,
         limit: int = 100,
+        ordering: List[OrderingItem] = None,
     ) -> List[CollectionWithStats]:
         return [
             self.from_stats_row(row)
             for row in (
                 self._items_with_stats.filter(*self.filter_criteria(pattern, libtype))
-                .order_by(Collection.name)
+                .order_by(*Collection.ordering_criteria(ordering))
```
MaciejWiczk, line 85: Should it be `Collection` or `CollectionWithStats`? Maybe this is causing an issue, when times_used on endpoint: `http://localhost:8000/api/v1/collections/stats/?limit=100&order=times_used` is not taken into account for sorting.

MaciejWiczk: Actually, now I read it second time, and times_used/avg_elapsed does not exist on Collection object/class definition, so I guess it won't be working that way

pbylicki: The idea here was to automatically allow sorting on all columns from database as for those we can safely assume that sorting is possible. Calculated columns/properties would need to be handled separately.

MaciejWiczk: Ok, so let's have a separate task for allowing sorting on additional columns/statistics, and merge that one.

MaciejWiczk (approved):

Issue #155, opened 2020-01-16
Library constructor is not documented


idefixcert:
I have a library which uses a constructor.
With libdoc the Doc of the constructor is also rendered in the document.
With rfhub2 not.

```python
class InfrastructurePathLibrary:
    """Library for Infrastructure Path.

    = Usage =

    This library helps to set the Python Path during the import time.

    """
    __version__ = '0.0.1'
    ROBOT_LIBRARY_SCOPE = 'TEST Suite'

    def __init__(self, path=None):
        """The library import:

        Examples:

        | =Setting= |     =Value=                       | =Value=                   |          =Comment=         |
        | Library   | InfrastructurePathLibrary  | path=../one               | # add one path             |
        | Library   | InfrastructurePathLibrary  | path=../one,/global       | # add two paths            |
        """

        # Set sys path parameters
        if path != None:
            path_list = path.split(",")
            for p in path_list:
                if p.startswith("/"):
                    sys.path.append(os.path.join(p))
                else:
                    sys.path.append(os.path.abspath(p))

    def nothing(self):
        pass
```

MaciejWiczk:
Yes, I have noticed it too. Currently the workaround I used is to document `__init__` either in library docs or in `__init__.py`.

Ps. Examples looks also nice when single `|` is used:
```
Examples:
        | =Setting=      =Value=                        =Value=                             =Comment=         
        | Library    InfrastructurePathLibrary   path=../one                # add one path             
        | Library    InfrastructurePathLibrary   path=../one,/global        # add two paths            
```
This will render docs like here:
![image](https://user-images.githubusercontent.com/35737166/72596365-2fa5b080-390c-11ea-84cf-c3bf0046d88e.png)

MaciejWiczk:
@idefixcert could You check #156 if that solves this issue, as well as #153 ?

idefixcert:
@MaciejWiczk yes with the latest release my issues are fixed

Issue #153, opened 2020-01-16
Importing a library which also contains robot files


idefixcert:
If I import a library which also contains robot files,
the robot files are not imported.

MaciejWiczk:
Hi @idefixcert, if I understood You correctly, You have library with __init__.py, and some .robot files in one directory? That's something I have not foreseen. Rfhub2, when traversing paths is looking for __init__.py file, and if it exists and importing it returns >0 keywords, this directory is treated as directory for single library, and the rest of the files is ommited. This way, app is not importing submodules, but Library as a whole.

idefixcert:
The problem is, there is no keyword in the library itself, I just want use the python setup tool for distributing the robot files via pip.
Is that a common way to do it?

MaciejWiczk:
> The problem is, there is no keyword in the library itself, I just want use the python setup tool for distributing the robot files via pip.

Oh, that's a tricky one. But is it safe to repoduce issue with assumptions I made previosly? I could change app to traverse path further for robot files, even where `__init__.py` exists. That could solve your issue.

idefixcert:
@MaciejWiczk yes with the latest release my issues are fixed

Issue #152, opened 2020-01-16
Print loaded collections when using CLI in alphabeticall order


MaciejWiczk:
Printing collections in alphabeticall order would help to visually indicate progress of long running command.

PR #156, opened 2020-01-17
Fixes for #152, #153, #155.

MaciejWiczk:
* #155, Extending Docs by init method docs, when init parametrised 
* #153, Allow importing robot files from directory with library with __init__.py
* #152, Printing loaded collections alphabetically


rfhub2/cli/rfhub_importer.py
```diff
@@ -238,6 +241,24 @@ def _serialise_keywords(self, libdoc: Dict) -> Dict:
                 keyword["args"] = ""
         return keywords
 
+    def _extract_doc_from_libdoc_inits(self, inits: List) -> str:
+        return "\n" + "\n" + "\n".join([d.doc for d in inits]) if len(inits) > 0 else ""
+
+    def _contains_robot_files(self, path: Path) -> bool:
+        return (
+            len(
+                [file for file in path.glob("**/*") if file.suffix in RESOURCE_PATTERNS]
```
pbylicki, line 250: it will be probably selecting multiple files that will be later discarded because it filters just based on the popular suffixes. It will get the job done as the files are later analyzed, but maybe it could be renamed to use "robot files candidates" or something similar that would indicate, that there may be wrong files identified.

tests/fixtures/updated/LibWithInit/test_res_lib_dir.resource
```diff
@@ -0,0 +1,14 @@
+*** Settings ***
```
pbylicki: it looks identical to the one from initial fixture directory. Is it necessary?

rfhub2/cli/rfhub_importer.py
MaciejWiczk: Goog idea, it's changed to that.

tests/fixtures/updated/LibWithInit/test_res_lib_dir.resource
MaciejWiczk: Deleted

pbylicki (approved):

PR #163, opened 2020-01-22
Fix click on keyword link and ellisis finding (#130)

MarcinMaciaszek:


MaciejWiczk:
@pbylicki @MarcinMaciaszek , what do You think aout bumping version to 0.13 so that we would release new version?

frontend/src/components/DrawerCollectionList.tsx
```diff
@@ -45,41 +45,66 @@ const DrawerCollectionListItem: React.FC<DrawerCollectionListItemProps> = observ
...
-      <ListItem
-        selected={isSelected(collection.id)}
-        onClick={event => handleListItemClick(event, collection.id)}
-      >
-        <ListItemText primary={collection.name} />
-        {isSelected(collection.id) ? <ExpandLess /> : <ExpandMore />}
-      </ListItem>
+      {(collection.name.length > 32) ? (
+        <Tooltip content={collection.name} direction="right" distance={26}>
+          <ListItem
```
pbylicki: This whole ListItem element could be defined once and turned into a function of collection and used here and in L64

MarcinMaciaszek: @pbylicki please check if I properly did it. I have tried in another way but without success :P

pbylicki: @MarcinMaciaszek I would split this new function into two:
`function collectionListItem(item: Collection): JSX.Element`
`function keywordListItem(item: Keyword): JSX.Element`
Otherwise it looks ok. You can also update version number as Maciej asked and it will be ready to merge :)

MarcinMaciaszek: Version bumped, will fix it according to your suggestions with the next ticket.

MaciejWiczk (changes requested):
build_ui.sh is failing. Need to investigate that

MaciejWiczk (approved):

pbylicki (approved):

Issue #31, opened 2019-08-19
Add handling for unauthenticated, deserialize and serialize in rfhub2.cli.api_client


MaciejWiczk:
Add handling for unauthenticated, deserialize and serialize in rfhub2.cli.api_client as described here:
[https://github.com/pbylicki/rfhub-new/pull/20#discussion_r314961693](url)

PR #168, opened 2020-01-27
Statistics extractor refactoring

pbylicki:
@MaciejWiczk Some refactoring to illustrate how we could use shared model between API and client code.

### DIFF
--- a/rfhub2/cli/statistics_importer.py
+++ b/rfhub2/cli/statistics_importer.py
@@ -1,8 +1,9 @@
 from pathlib import Path
-from typing import Dict, List, Set, Tuple
+from typing import List, Set, Tuple
 
 from .api_client import Client
 from .statistics_extractor import StatisticsExtractor
+from rfhub2.model import KeywordStatistics
 
 
 class StatisticsImporter:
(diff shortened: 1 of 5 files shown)


rfhub2/cli/statistics_importer.py
```diff
@@ -1,82 +1,83 @@
...
-        This is simplified approach to quick check file.
-        :param path:
-        :return:
-        """
-        with open(path, "r", encoding="utf-8", errors="ignore") as f:
-            # read the first few lines; if we don't see
-            # what looks like robot generator tag, return false
-            data = f.read(200)
-            return "<robot generator=" in data.lower()
+from pathlib import Path
```
MaciejWiczk, line 1: Anything happened to this file or is it just Win/Lin conversion?

MaciejWiczk (approved):
image

MaciejWiczk:
> @MaciejWiczk Some refactoring to illustrate how we could use shared model between API and client code.

Is it part of #31?

rfhub2/cli/statistics_importer.py
pbylicki: only meaningful change is updated type annotation in L48

pbylicki:
> > @MaciejWiczk Some refactoring to illustrate how we could use shared model between API and client code.
> 
> Is it part of #31?

It is a step in that direction :) #31 will require doing this for keywords and collections with some refactorings available when all components follow the same approach.

MaciejWiczk:
> > > @MaciejWiczk Some refactoring to illustrate how we could use shared model between API and client code.
> > 
> > 
> > Is it part of #31?
> 
> It is a step in that direction :) #31 will require doing this for keywords and collections with some refactorings available when all components follow the same approach.

I will create PR with some changes for keywords and collections, now that I know how it should be done.

rfhub2/cli/statistics_importer.py
MaciejWiczk: GitHub was just showing that all file has changed. Strange

PR #180, opened 2020-02-11
Change statistics model and endpoints typing

MaciejWiczk:
This is a PR for initial review. I do not have time to refactor tests, but they do deserve that. A lot.

### DIFF
--- a/rfhub2/cli/keywords_importer.py
+++ b/rfhub2/cli/keywords_importer.py
@@ -3,6 +3,7 @@ from pathlib import Path
 import re
 from robot.errors import DataError
 from robot.libdocpkg import LibraryDocumentation
+from robot.libdocpkg.model import LibraryDoc
 import robot.libraries
 from typing import Dict, List, Set, Tuple
 
@@ -10,7 +11,6 @@ from .api_client import Client
 from rfhub2.model import (
     Collection,
     CollectionUpdate,
-    Keyword,
     KeywordCreate,
     KeywordUpdate,
     NestedKeyword,
@@ -234,37 +234,35 @@ class KeywordsImporter(object):
             )
         return loaded_collections
 
-    def _serialise_libdoc(self, libdoc: Dict, path: str) -> CollectionUpdate:
+    def _serialise_libdoc(self, libdoc: LibraryDoc, path: str) -> CollectionUpdate:
         """
-        Serialises libdoc object to Collection object.
-        :param libdoc: libdoc input object
+        Serialises LibraryDoc object to CollectionUpdate object.
+        :param libdoc: LibraryDoc input object
         :param path: library path
         :return: CollectionUpdate object
         """
-        lib_dict = libdoc.__dict__
         return CollectionUpdate(
-            name=lib_dict["name"],
-            type=lib_dict["type"],
-            version=lib_dict["version"],
-            scope=lib_dict["scope"],
-            # named_args=lib_dict["named_args"], # we have not used this one, yet
+            name=libdoc.name,
+            type=libdoc.type,
+            version=libdoc.version,
+            scope=libdoc.scope,
+            # named_args=libdoc.named_args, # we have not used this one, yet
             path=path,
-            doc=lib_dict["doc"]
-            + self._extract_doc_from_libdoc_inits(lib_dict["inits"]),
-            doc_format=lib_dict["_setter__doc_format"],
+            doc=libdoc.doc + self._extract_doc_from_libdoc_inits(libdoc.inits),
+            doc_format=libdoc._setter__doc_format,
         )
 
-    def _serialise_keywords(self, libdoc: Dict) -> List[KeywordUpdate]:
+    def _serialise_keywords(self, libdoc: LibraryDoc) -> List[KeywordUpdate]:
         """
-        Serialises keywords to Keyword object.
-        :param :libdoc input object
+        Serialises keywords to KeywordUpdate object.
+        :param :LibraryDoc input object
         :return: KeywordUpdate object
         """
         return [
             KeywordUpdate(
-                name=keyword.__dict__["name"],
-                args=self._serialise_args(keyword.__dict__["args"]),
-                doc=keyword.__dict__["doc"],
+                name=keyword.name,
+                args=self._serialise_args(keyword.args),
+                doc=keyword.doc,
             )
             for keyword in libdoc.keywords
         ]
--- a/rfhub2/cli/keywords_importer.py
+++ b/rfhub2/cli/keywords_importer.py
@@ -160,9 +160,7 @@ class KeywordsImporter(object):
                 )
         return sorted(collections, key=lambda i: i[0].name)
 
-    def create_collection(
-        self, path: Path
-    ) -> CollectionUpdateWithKeywords:
+    def create_collection(self, path: Path) -> CollectionUpdateWithKeywords:
         """
         Creates Collection object from provided path.
         :param path: Path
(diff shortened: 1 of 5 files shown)


rfhub2/cli/keywords_importer.py
```diff
@@ -129,7 +133,9 @@ def _traverse_paths(self, path: Path) -> Set[Path]:
                     valid_lib_paths.add(item)
         return valid_lib_paths
 
-    def create_collections(self, paths: Set[Path]) -> List[Dict]:
+    def create_collections(
+        self, paths: Set[Path]
+    ) -> List[Tuple[Collection, List[KeywordUpdate]]]:
```
pbylicki: This tuple type appears a lot in the logic. Maybe you could define a dataclass with these two fields so you have clearer signatures and can access named fields instead of indexed items of a tuple

MaciejWiczk: Should I put dataclass in `rfhub2/db/repository/keyword_statistics_repository.py` or somewhere else?

pbylicki: @MaciejWiczk as far as I can see it will be used only in this module so you can add it here

MaciejWiczk: Done. I cherry-picked commit for fixing acceptance tests, to check pipeline.

rfhub2/cli/keywords_importer.py
```diff
@@ -138,27 +151,29 @@ def create_collections(self, paths: Set[Path]) -> List[Dict]:
...
+    def create_collection(self, path: Path) -> CollectionUpdateWithKeywords:
         """
         Creates Collection object from provided path.
         :param path: Path
         :return: Collection object
         """
         libdoc = LibraryDocumentation(str(path))
         serialised_keywords = self._serialise_keywords(libdoc)
-        return self._serialise_libdoc(libdoc, str(path), serialised_keywords)
+        return self._serialise_libdoc(libdoc, str(path)), serialised_keywords
```
pbylicki: tuple returned, not dataclass instance

rfhub2/cli/keywords_importer.py
```diff
@@ -185,67 +202,78 @@ def delete_outdated_collections(
...
+        )
+
+    def _serialise_keywords(self, libdoc: Dict) -> List[KeywordUpdate]:
         """
         Serialises keywords to Keyword object.
         :param :libdoc input object
-        :return: Keyword object
+        :return: KeywordUpdate object
         """
+        raw_keywords = [keyword.__dict__ for keyword in libdoc.keywords]
```
pbylicki: Could you just create KeywordUpdate from keyword attributes in first list comprehension?

rfhub2/cli/keywords_importer.py
```diff
@@ -346,144 +374,135 @@ def _has_test_case_table(data: str) -> bool:
...
     @staticmethod
     def _get_obsolete_collections_ids(
-        existing_collections: List[Dict], new_collections: List[Dict]
+        existing_collections: List[Collection],
+        new_collections: List[CollectionUpdateWithKeywords],
     ) -> Set[int]:
         """Returns set of collection ids that were found in application but not in paths"""
         new_collections_paths = {
-            new_collection["path"] for new_collection in new_collections
+            new_collection[0].path for new_collection in new_collections
```
pbylicki: Type annotation for new collections does not match actual type. There is CollectionUpdateWithKeywords dataclass in annotation and the code is using a tuple. The same happens also in the following methods

rfhub2/cli/keywords_importer.py
```diff
@@ -21,6 +31,12 @@
 INIT_FILES = {"__init__.txt", "__init__.robot", "__init__.html", "__init__.tsv"}
 
 
+@dataclass
+class CollectionUpdateWithKeywords:
+    collection: CollectionUpdate
```
MaciejWiczk, line 36: Should it be declared this way:
```
class CollectionUpdateWithKeywords:
    data: Tuple[CollectionUpdate, List[KeywordUpdate]]

File: rfhub2/cli/keywords_importer.py [hunk 6]
```diff
@@ -185,67 +202,78 @@ def delete_outdated_collections(
             self.client.delete_collection(collection)
         return collections_to_delete
 
-    def add_collections(self, collections: List[Dict]) -> List[Dict[str, int]]:
+    def add_collections(
+        self, collections: Tuple[List[CollectionUpdate], List[KeywordUpdate]]
```
MaciejWiczk: I need to check this annotation as well. BTW, maybe validator like MyPy or smth similar could be added to CI?

rfhub2/cli/keywords_importer.py
MaciejWiczk: Or should I update the way the functions does return items as instance of this class?

pbylicki: > Or should I update the way the functions does return items as instance of this class?

this one would be better

rfhub2/cli/keywords_importer.py
pbylicki: We can investigate this. In the meantime, PyCharm is warning about such discrepancies between annotation and code pretty well

pbylicki: @MaciejWiczk Once we merge our PRs I will work on adding mypy check and resolving any typing issues. There are typing stubs for SQLAlchemy that will hopefully work. As far as I know there are none for robot framework so part of the code will be unchecked. But still I really like this idea and want to introduce it.

MaciejWiczk: >  Once we merge our PRs I will work on adding mypy check and resolving any typing issues.

Good, I feel like I could made a few mistakes here and there.

> There are typing stubs for SQLAlchemy that will hopefully work.

You mean to make MyPy not go crazy with SQLAlchemy?

> As far as I know there are none for robot framework so part of the code will be unchecked.

You mean there is no stubs for RFWK? Maybe we could exclude `*.robot, *.resource`  files alltogether?

Here is a [task for MyPy](https://github.com/pbylicki/rfhub2/issues/196)

pbylicki: MyPy is based on type annotations which are still not available for some Python packages, like SQLAlchemy or Robot Framework. For SQLAlchemy there is external package with stubs maintained by Dropbox, but for Robot Framework there is nothing yet. So any code in our project that interacts with imports from robot package won't be properly checked.

rfhub2/cli/keywords_importer.py
MaciejWiczk: Could You take a look now? I don't exaclty like the look of it now, however it's in one statement.

rfhub2/cli/keywords_importer.py
```diff
@@ -185,67 +203,80 @@ def delete_outdated_collections(
...
+                }
             )
             print(
-                f'{collection["name"]} library with {len(collection["keywords"])} keywords loaded.'
+                f"{collection.collection.name} library with {len(collection.keywords)} keywords loaded."
             )
         return loaded_collections
 
-    def _serialise_libdoc(self, libdoc: Dict, path: str, keywords: Dict) -> Dict:
+    def _serialise_libdoc(self, libdoc: Dict, path: str) -> CollectionUpdate:
```
MaciejWiczk: I left this one unchanged, but I know it's not right. Should I use `LibraryDocumentation` class as type for `libdoc` instead of Dict?

pbylicki: libdoc returned by LibraryDocumentation factory method is instance of libraryDoc class:
`from robot.libdocpkg.model import LibraryDoc`

MaciejWiczk: Fixed, thanks for help! :) I think we are ready to to merge it into main statistic branch?

pbylicki (approved):

PR #191, opened 2020-03-14
Fix cli tests

pbylicki:
Changes in Click 7.1 broke robot CLI tests that relied on specific line wrapping of help message. Added stripping help output of new lines/ indents to make it more robust


MaciejWiczk (approved):
I was just about to do the same :)

PR #195, opened 2020-03-16
Add many statistics endpoint

pbylicki:
Updated existing endpoint to accept multiple records. Updated CLI to send payloads as one item list for now. Importer should be updated to upload records in batches

### DIFF
--- a/rfhub2/model/__init__.py
+++ b/rfhub2/model/__init__.py
@@ -88,3 +88,7 @@ class KeywordStatistics(BaseModel):
 
 class StatisticsDeleted(BaseModel):
     deleted: int
+
+
+class StatisticsInserted(BaseModel):
+    inserted: int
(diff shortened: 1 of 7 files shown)


pbylicki:
@MaciejWiczk Tests fail because of this Click issue. I will wait for your changes with a fix to be merged then.

rfhub2/model/__init__.py
```diff
@@ -86,5 +86,21 @@ class Config:
...
+
+    @staticmethod
+    def of(items: List[KeywordStatistics]) -> "KeywordStatisticsList":
+        return KeywordStatisticsList(__root__=items)
+
+    @staticmethod
+    def one(item: KeywordStatistics) -> "KeywordStatisticsList":
+        return KeywordStatisticsList(__root__=[item])
+
+
```
MaciejWiczk, line 100: What do those methods do?

MaciejWiczk (approved):

rfhub2/model/__init__.py
pbylicki: I wanted to add such helpers so we do not need to explicitly instantiate this custom root type everywhere: https://pydantic-docs.helpmanual.io/usage/models/#custom-root-types

PR #200, opened 2020-03-24
Modifying statistics_importert to send stats in batch per file

### LINKED ISSUE #197: Update CLI for statistics to work in batch mode
Currently statistics are working in one by one mode, which takes a lot of time. Switching to batch mode, will shorten execution time significantly.

MaciejWiczk:
This addresses #197.
I've added progress bar to indicate that app is processing requests, since it took a while on my corporate network, maybe due to some high network usage lately. Here are some benchmarks:
image

So network lag was reduced to minimum with batch mode. During spike usage of RAM was about 1 GB, but that happened when processing 700 MB file. Normally usage is around 150 MB for 1.5 GB bunch of logs.

### DIFF
--- a/rfhub2/cli/statistics_importer.py
+++ b/rfhub2/cli/statistics_importer.py
@@ -26,11 +26,20 @@ class StatisticsImporter:
         :return: Number of libraries and keyword loaded
         """
         execution_files = self.get_execution_files_paths()
-        statistics = [
-            StatisticsExtractor(execution_file).compute_statistics()
-            for execution_file in execution_files
-        ]
-        return self.add_statistics(statistics)
+        loaded_statistics = []
+        progress_bar = IncrementalBar(
+            "Sending statistics",
+            max=len(execution_files),
+            suffix="%(percent).1f%% - %(eta)ds, elapsed: %(elapsed)ds",
+        )
+        for execution_file in execution_files:
+            statistics = StatisticsExtractor(execution_file).compute_statistics()
+            loaded_statistics.append(self.add_statistics(statistics, execution_file))
+            progress_bar.next()
+        return (
+            len([stat for stat in loaded_statistics if stat > 0]),
+            sum(loaded_statistics),
+        )
 
     def get_execution_files_paths(self) -> Set[Path]:
         """
@@ -46,35 +55,22 @@ class StatisticsImporter:
         }
 
     def add_statistics(
-        self, statistics: List[List[KeywordStatistics]]
-    ) -> Tuple[int, int]:
+        self, statistics: List[KeywordStatistics], execution_file: Path
+    ) -> int:
         """
         Adds statistics from provided list to app.
         :param statistics: List of KeywordStatistics objects
-        :return: list of dictionaries with collection name and number of keywords.
+        :param execution_file: Path to file from statistics where extracted
+        :return: number of statistics sent to app
         """
-        collections, keywords = set(), set()
-        progress_bar = IncrementalBar(
-            "Sending statistics",
-            max=len(statistics),
-            suffix="%(percent).1f%% - %(eta)ds, elapsed: %(elapsed)ds",
-        )
-        for stats in statistics:
-            stats_req = self.client.add_statistics(KeywordStatisticsList.of(stats))
-            if stats_req[0] == 201:
-                collections.update({stat.collection for stat in stats})
-                keywords.update(
-                    {".".join((stat.collection, stat.keyword)) for stat in stats}
-                )
-                progress_bar.next()
-            elif stats_req[0] != 400:
-                print(stats_req[1]["detail"])
-                raise StopIteration
-            else:
-                print(
-                    f"""Record already exists for provided execution_time: {stats[0].execution_time}"""
-                )
-        return len(collections), len(keywords)
+        stats_req = self.client.add_statistics(KeywordStatisticsList.of(statistics))
+        if stats_req[0] == 201:
+            return len(statistics)
+        elif stats_req[0] != 400:
+            print(stats_req[1]["detail"])
+        else:
+            print(f"""Records already exist for file from {execution_file}""")
+        return 0
 
     @staticmethod
     def _is_valid_execution_file(path: Path) -> bool:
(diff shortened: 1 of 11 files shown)


rfhub2/cli/statistics_importer.py
```diff
@@ -45,27 +45,35 @@ def get_execution_files_paths(self) -> Set[Path]:
             if self._is_valid_execution_file(p)
         }
 
-    def add_statistics(self, statistics: List[KeywordStatistics]) -> Tuple[int, int]:
+    def add_statistics(
+        self, statistics: List[KeywordStatisticsList]
```
pbylicki: statistics type is List[List[KeywordStatistics]]

rfhub2/cli/statistics_importer.py
```diff
@@ -45,27 +45,35 @@ def get_execution_files_paths(self) -> Set[Path]:
...
+                print(stats_req[1]["detail"])
                 raise StopIteration
             else:
                 print(
-                    f"""Record already exists for provided collection: {
-                    stat.collection}, keyword: {stat.keyword} and execution_time: {stat.execution_time}"""
+                    f"""Record already exists for provided execution_time: {stats[0].execution_time}"""
                 )
-
         return len(collections), len(keywords)
```
pbylicki: Do you think that this return information is required? The method is uploading statistics records for multiple files and we are not showing anywhere how much statistics records we uploaded, but only the number of unique collections and keywords across all the files. It seems like a secondary information. Maybe you could also present the number of statistics uploaded? If you upload N files with different executions of the same suite, the number of collections and keywords may not reflect the number of statistics uploaded which can be up to N*keywords

rfhub2/cli/statistics_importer.py
```diff
@@ -45,27 +45,35 @@ def get_execution_files_paths(self) -> Set[Path]:
...
+                )
+                progress_bar.next()
+            elif stats_req[0] != 400:
+                print(stats_req[1]["detail"])
                 raise StopIteration
             else:
                 print(
-                    f"""Record already exists for provided collection: {
-                    stat.collection}, keyword: {stat.keyword} and execution_time: {stat.execution_time}"""
+                    f"""Record already exists for provided execution_time: {stats[0].execution_time}"""
```
pbylicki: maybe this message would be more helpful if it shows which file statistics failed to upload

rfhub2/cli/statistics_importer.py
MaciejWiczk: Yes, You are totally right. I could change it according to your comment. I wanted to keep both importers to work the same way and return `Tuple[int, int]` that is used [here](https://github.com/pbylicki/rfhub2/blob/master/rfhub2/cli/cli.py#L64). But still, I can add total number of collections and keywords, instead of unique ones. Is that alright?

rfhub2/cli/statistics_importer.py
MaciejWiczk: Definetely, it would require to change statistics_extractor.py accordingly, to keep that info.

MaciejWiczk: How about making KeywordStatistics class hold that info?

pbylicki: how about making this method responsible for upload of single file statistics. You could move progressbar and results aggregation to parent method and pass file as additional argument to this method

rfhub2/cli/statistics_importer.py
pbylicki: I think it would be good to at least log the total number of statistic records uploaded. If you want to keep the numbers of collections and keywords they should be unique to not produce misleading message. After all you are counting here references to collections and keywords, not these objects itself

MaciejWiczk: I will change the message to files and statistics number, that would be clearer.

pbylicki: Good idea!

MaciejWiczk:
@pbylicki how about this one PR? Is it ok, or should I still change something?

rfhub2/cli/statistics_importer.py
```diff
@@ -45,28 +54,23 @@ def get_execution_files_paths(self) -> Set[Path]:
...
-                )
-
-        return len(collections), len(keywords)
+        stats_req = self.client.add_statistics(KeywordStatisticsList.of(statistics))
+        if stats_req[0] == 201:
+            return len(statistics)
+        elif stats_req[0] != 400:
+            print(stats_req[1]["detail"])
+        else:
+            print(f"""Records already exist for file from {execution_file}""")
```
pbylicki, line 72: You could add unit test for this branch, then the PR would not decrease coverage. You can also remove `KeywordStatisticsList.one` method because it is substituted everywhere with `KeywordStatisticsList.of` method now.

MaciejWiczk: Done. Coverage results are impacted mostly by cli.py, which length has increased. This one is covered fully by accptance tests, so maybe. we could exclude it wih proper annotation?

pbylicki: Thanks, I know, but still wanted to improve it where it is easy to do. I would prefer to add the unit tests for cli in the future instead.

pbylicki:
@MaciejWiczk looks like acceptance tests are failing now

MaciejWiczk:
> @MaciejWiczk looks like acceptance tests are failing now

Yes, I'm investigating, because they work locally perfeclty fine.

rfhub2/cli/statistics_importer.py
MaciejWiczk: I'm all for it. Last time I struggled with mocking cli, but maybe there is some proper way to do it, that I do not know of.

MaciejWiczk:
> > @MaciejWiczk looks like acceptance tests are failing now
> 
> Yes, I'm investigating, because they work locally perfeclty fine.

I'm totally puzzled. Selenium nows does not recognize locator. It's not that my branch is failing, but all of the branches are failing, when build is rerun. @pbylicki Do You have any idea? I updated selenium WebDriver to match chrome version, but it didn't help.

pbylicki (approved):

PR #201, opened 2020-03-26
Making add_statistics into method for single list not list of lists

MaciejWiczk:
@pbylicki, I came up with such changes. I wanted to check them on corp network, thus this is a separate PR.


MaciejWiczk:
setup.py should be updated to require Fastapi and pydantic in higher versions. Installing it now result in fail with statistics list. I wil find minimum version later.

Edit: fastapi >= 0.46 solves the issue with pydantic when it's not installed. I've added pydantic abov 1.0 when existing installation of 0.32.2 is found.

pbylicki (approved):

PR #203, opened 2020-03-29
Feature/rfhub2 161 add tags

### LINKED ISSUE #158: Extend keyword model so tags can be gathered
Child issue of #136.
Model would need to be extended with additional, optional field of list type, similar to args. This needs to happen for:

- [DB](https://github.com/pbylicki/rfhub2/blob/master/rfhub2/db/model/keyword.py#L13)
- [API](https://github.com/pbylicki/rfhub2/blob/master/rfhub2/model/__init__.py#L25)

### LINKED ISSUE #159: Search endpoint should look in tags as well
Child issue of #136.
When using search endpoint it should scan existing tags on keyword level. It should also allow for searching tags only using `tag: tag` similar to current `name: name` `in: in` with searchbox on app.
This issue depends on #158. Code should be changed in:

- [Search Params](https://github.com/pbylicki/rfhub2/blob/4b0475048c27e9d2d5a40404d62af01a85d75423/rfhub2/ui/search_params.py#L21)

### LINKED ISSUE #160: CLI should gather tags from keywords
Child issue of #136.
Currently CLI do not gather tags, but discards them. This is a simple change, that can be done afer #158. Code to change:

- [rfhub_importer](https://github.com/pbylicki/rfhub2/blob/4b0475048c27e9d2d5a40404d62af01a85d75423/rfhub2/cli/rfhub_importer.py#L232)

### LINKED ISSUE #161: Frontend should display tags
Child issue of #136.
After #158, #159, #160 frontend should be changed to display tags values for keywords. Code to change:

- [Collection Details](https://github.com/pbylicki/rfhub2/blob/4b0475048c27e9d2d5a40404d62af01a85d75423/frontend/src/components/CollectionDetails.tsx#L34)
- [Search Results](https://github.com/pbylicki/rfhub2/blob/4b0475048c27e9d2d5a40404d62af01a85d75423/frontend/src/components/SearchKeywordList.tsx#L37)

MaciejWiczk:
Adding tags to frontend and backend/cli. No alembic migrations are prepared yet, I will need to figure it out.


MaciejWiczk:
I will add missing tests in nearest future

MaciejWiczk:
@pbylicki I took a look at alembic and it requires your expertise. There is a bit of design stuff, before coding. If migration is to be invoked programatically, there is a need for alembic.ini/env.py getting env variables, then running it alongside app etc. Could You take a look at this stuff?

MaciejWiczk:
@pbylicki could You highlight in code where should I do changes to return tags as list? And how should I save values to DB for tags? I went through the docs, but I still don't get it.

rfhub2/cli/keywords_importer.py
```diff
@@ -262,6 +262,7 @@ def _serialise_keywords(self, libdoc: LibraryDoc) -> List[KeywordUpdate]:
             KeywordUpdate(
                 name=keyword.name,
                 args=self._serialise_args(keyword.args),
+                tags=self._serialise_args(keyword.tags),
```
MaciejWiczk: `tag=keyword.tags`?

rfhub2/db/model/mixins.py
```diff
@@ -19,3 +19,10 @@ def arg_string(self) -> str:
         Old implementation saves args list as JSON in text field, this is more readable representation for UI
         """
         return ", ".join(json.loads(self.args)) if self.args else ""
+
+    @property
+    def tags_string(self) -> str:
```
MaciejWiczk: this one should be removed. Should something be added instead of this one for tags?

rfhub2/model/__init__.py
```diff
@@ -27,6 +27,7 @@ class KeywordUpdate(BaseModel):
     name: str
     doc: Optional[str]
     args: Optional[str]
+    tags: Optional[str]
```
MaciejWiczk: Should it be: `tags: List[Optional[str]] = []`

rfhub2/model/__init__.py
```diff
@@ -52,6 +53,7 @@ class NestedKeyword(KeywordUpdate):
     synopsis: Optional[str]
     html_doc: Optional[str]
     arg_string: Optional[str]
+    tags_string: Optional[str]
```
MaciejWiczk: This one should be deleted

rfhub2/db/model/keyword.py
```diff
@@ -11,12 +11,13 @@ class Keyword(Base, KeywordMixin):
     name = Column(Text, index=True)
     doc = Column(Text)
     args = Column(Text)
+    tags = Column(Text)
```
MaciejWiczk, line 16: I'm not sure about this one, what kind of changes should happen to properly save tags? Should the be exaclty like args?

pbylicki:
> @pbylicki could You highlight in code where should I do changes to return tags as list? And how should I save values to DB for tags? I went through the docs, but I still don't get it.

I think we should have tags represented as list of strings on the API schema level and as a string with JSON list in the DB (like you did). In this way we will have validation of this field instead of just hoping that users will always upload a valid list there. SQLite does not have support for array columns so we would need to drop it and only use Postgres from now. Once this is done we should actually do the same for args field to add validation there and have consistent schema

frontend/src/types/ModelTypes.ts
```diff
@@ -3,7 +3,9 @@ export interface Keyword {
     name: string
     doc: string
     args: string | null
+    tags: string | null
```
pbylicki: should be `string[]` instead, tags_string could be turned into method using tags field

frontend/src/types/ModelTypes.ts
```diff
@@ -38,6 +42,10 @@ export interface Collection {
 export interface NestedCollection {
     id: number
     name: string
+    args: string | null
```
pbylicki: Should collection have these fields too?

rfhub2/db/model/mixins.py
```diff
@@ -19,3 +19,10 @@ def arg_string(self) -> str:
         Old implementation saves args list as JSON in text field, this is more readable representation for UI
         """
         return ", ".join(json.loads(self.args)) if self.args else ""
+
+    @property
```
pbylicki: IMO this can be removed from API and put  in UI if it will be used after all (chips?)

rfhub2/model/__init__.py
pbylicki: tags: List[str]

frontend/src/types/ModelTypes.ts
MaciejWiczk: ok, should it be everywhere instead of string | null?

rfhub2/db/model/mixins.py
MaciejWiczk: UI? How is it possible? I don't see that in ui folder

pbylicki:
@MaciejWiczk One missing element is two-way mapping between pydantic model and DB model which will need to be explicit as we introduce custom mapping between the fields. I wanted to do this soon because lack of this is actually the source of the problem with latest FastAPI version. I will try to add it today/tomorrow to not keep you blocked.

MaciejWiczk (commented):
> @MaciejWiczk One missing element is two-way mapping between pydantic model and DB model which will need to be explicit as we introduce custom mapping between the fields. I wanted to do this soon because lack of this is actually the source of the problem with latest FastAPI version. I will try to add it today/tomorrow to not keep you blocked.

@pbylicki, I will not be able to introduce this changes. TBH I do not get fastapi framework. If there are some not breaking changes I can handle it, but the changes You mentioned are out of my abilities.

rfhub2/db/repository/keyword_repository.py
```diff
@@ -146,3 +148,8 @@ def get_with_stats(self, item_id: int) -> Optional[KeywordWithStats]:
         result = self._items_with_stats.filter(self._id_filter(item_id)).first()
         if result:
             return self.from_stats_row(result)
+
+    def update(self, item: Keyword, update_data: dict) -> Keyword:
+        if 'tags' in update_data:
+            update_data['tags'] = Keyword.from_json_list(update_data['tags'])
+        super().update(item, update_data)
```
pbylicki: `return` is required to return Keyword

rfhub2/templates/index.html
```diff
@@ -1 +1 @@
...
...1:0},c=[];function f(e){if(t[e])return t[e].exports;var r=t[e]={i:e,l:!1,exports:{}};return i[e].call(r.exports,r,r.exports,f),r.l=!0,r.exports}f.m=i,f.c=t,f.d=function(e,r,t){f.o(e,r)||Object.defineProperty(e,r,{enumerable:!0,get:t})},f.r=function(e){"undefined"!=typeof Symbol&&Symbol.toStringTag&&Object.defineProperty(e,Symbol.toStringTag,{value:"Module"}),Object.defineProperty(e,"__esModule",{value:!0})},f.t=function(r,e){if(1&e&&(r=f(r)),8&e)return r;if(4&e&&"object"==typeof r&&r&&r.__esModule)return r;var t=Object.create(null);if(f.r(t),Object.defineProperty(t,"default",{enumerable:!0,value:r}),2&e&&"string"!=typeof r)for(var n in r)f.d(t,n,function(e){return r[e]}.bind(null,n));return t},f.n=function(e){var r=e&&e.__esModule?function(){return e.default}:function(){return e};return f.d(r,"a",r),r},f.o=function(e,r){return Object.prototype.hasOwnProperty.call(e,r)},f.p="/static/";var r=window.webpackJsonpfrontend=window.webpackJsonpfrontend||[],n=r.push.bind(r);r.push=e,r=r.slice();for(var o=0;o<r.length;o++)e(r[o]);var s=n;a()}([])</script><script src="/static/static/js/2.371e99a6.chunk.js"></script><script src="/static/static/js/main.c07b13a9.chunk.js"></script></body></html>
```
pbylicki: This file shouldn't be updated

MaciejWiczk: fixed, added by mistake

rfhub2/db/repository/keyword_repository.py
MaciejWiczk: like `return super().update(item, update_data)` ?

tests/unit/db/base_repo_tests.py
```diff
@@ -22,14 +24,20 @@ def setUp(self) -> None:
...
+            Keyword(name="Login keyword", doc="Perform some check", tags=['tag_1', 'tag_2']),
+            Keyword(name="Teardown", doc="Clean up environment", tags=['tag_1', 'tag_2']),
         ]
         self.app_keyword = Keyword(name="Login to Application")
 
+        kwd_tmp = deepcopy(self.keywords)
+        for keyword in kwd_tmp:
+            if keyword.tags:
+                keyword.tags = json.dumps(keyword.tags)
+
```
MaciejWiczk: @pbylicki this one is bugging me. I have no idea why this is failing. If I leave raw lists of straings they are not properly transported to db (insert for list lack '' around []). Could You help me on that one?

rfhub2/cli/keywords_importer.py
```diff
@@ -276,6 +278,9 @@ def _serialise_args(self, args: List[str]) -> str:
             else ""
         )
 
+    def _serialise_tags(self, tags) -> str:
```
pbylicki: what is the type of single tag here? if it is `str` than `str()` is not required. If it some class instance, maybe you could access some field to get tag name.
Return type should be `List[str]`

MaciejWiczk: I will check that and introduce proper type annotation

MaciejWiczk: It would be changed to:
```
from robot.model import tags

    def _serialise_tags(self, tags: tags) -> List[str]:
        return list(tags._tags)

File: tests/unit/db/base_repo_tests.py (hunk 11 above)
Comment (pbylicki) [reply tests/unit/db/base_repo_tests.py]: Keyword instances here are instances of database model and require `tags` as optional string

File: rfhub2/cli/keywords_importer.py [hunk 13]
```diff
@@ -276,6 +278,9 @@ def _serialise_args(self, args: List[str]) -> str:
             else ""
         )
 
+    def _serialise_tags(self, tags: tags) -> List[str]:
```
pbylicki: @MaciejWiczk  tags is a module. There is a `Tags` class inside which should be used here

tests/acceptance/e2e.robot
```diff
@@ -76,19 +76,40 @@ Search Should Return Expected Results
...
-    name:Sh in:Te   1    Someone Shall Pass    ${EMPTY}    Test Libdoc File
+    Method 3    1    Single Class Lib Method 3    ${EMPTY}    SingleClassLib    Docstring for single_class_lib_method_3 with two params
+    Keyword*Doub    1    Keyword With Args With Double Quotation Mark    ${EMPTY}    test_robot    Keyword With Args With Double Quotation Mark
+    name:Some    1    Someone Shall Pass    ${EMPTY}    ${EMPTY}    Test Libdoc File
+    Some in:Te    1   Someone Shall Pass    ${EMPTY}    ${EMPTY}    Test Libdoc File
+    name:Sh in:Te   1    Someone Shall Pass    ${EMPTY}    ${EMPTY}    Test Libdoc File
+    tags:tag_1   1    Single Class Lib Method 1    ${EMPTY}    SingleClassLib    Docstring for single_class_lib_method_1
+    tags:tag in:Single   1    Single Class Lib Method 1    ${EMPTY}    SingleClassLib    Docstring for single_class_lib_method_1
+
+Tags Should Be Dispalyed On Collection Details
```
pbylicki: spelling error in displayed

tests/acceptance/e2e.robot
```diff
@@ -76,19 +76,40 @@ Search Should Return Expected Results
...
+    [Documentation]    This test bases on 
+    ...    'Populated App Should Show Number Of Collections'
+    ...    to shorten execution time.
+    [Tags]    rfhub2-161    tags
+    [Template]    Tags Should Be Displayed For Collection
+    test_robot        first_tag    second_tag
+    SingleClassLib    tag_1        tag_2
+
+Tags Should Be Dispalyed On Search Results
```
pbylicki: spelling error in displayed

MaciejWiczk:
@pbylicki there is some strange behaviour from TravisCI. It stalls on robot tests. @MarcinMaciaszek will investigate GitHub Actions and we will see how it will behave there.

rfhub2/cli/keywords_importer.py
MaciejWiczk: @pbylicki, it should be alright now. Unfortunately this class does not seem to have string representation, but using `list(tags._tags)` it works. If I missed best solution please let me know.

tests/acceptance/e2e.robot
MaciejWiczk: fixed

tests/acceptance/e2e.robot
MaciejWiczk: fixed

MaciejWiczk:
> @pbylicki there is some strange behaviour from TravisCI. It stalls on robot tests. @MarcinMaciaszek will investigate GitHub Actions and we will see how it will behave there.

I did a workaround to run tests for CLI nad E2E separately. Ugly, but works for now. In a separate PR I will add tests for statistics, because they failed the same way.

rfhub2/cli/keywords_importer.py
pbylicki: that's ok, I just meant that `tags` parameter is of class `Tags`, not `tags`

pbylicki (approved):

PR #208, opened 2020-03-30
Add explicit model mapping

pbylicki:


rfhub2/db/model/keyword.py
```diff
@@ -24,3 +25,36 @@ def __repr__(self):  # pragma: no cover
     @classmethod
     def default_ordering(cls) -> List[OrderingItem]:
         return [OrderingItem("name")]
+
+    @staticmethod
+    def create(data: model.KeywordCreate) -> "Keyword":
+        return Keyword(
+            name=data.name,
+            doc=data.doc,
+            args=data.args,
```
pbylicki, line 34: @MaciejWiczk you could add here conversion of `tags: List[str]` from `data` to `tags: str` in DB model by using json serialization

rfhub2/db/model/keyword.py
```diff
@@ -24,3 +25,36 @@ def __repr__(self):  # pragma: no cover
...
+            collection_id=data.collection_id,
+        )
+
+    def to_model(self) -> model.Keyword:
+        collection = self.collection
+        return model.Keyword(
+            id=self.id,
+            name=self.name,
+            doc=self.doc,
+            args=self.args,
```
pbylicki, line 44: @MaciejWiczk you could add here conversion of `tags: str` in DB model to `tags: List[str]` in `mode.Keyword` by using json deserialization

rfhub2/db/model/keyword.py
MaciejWiczk: What do You mean by json serialization? json.loads and then some list comprehension or something totally different? Or using json encoder like [here](https://fastapi.tiangolo.com/tutorial/encoder/) just on tags object?

MaciejWiczk:
Before merging this one we should update version.py. If those changes are not be backward compatible.

pbylicki:
> Before merging this one we should update version.py. If those changes are not be backward compatible.

Could you explain? These changes fix wrong behavior of some of the endpoints that used to work with old versions of fastapi, but don't work with latest ones. There is no change in the API schema.
My goal was to merge it so you can use it in your MR with tags so that they can be released together.

MaciejWiczk (approved):
> > Before merging this one we should update version.py. If those changes are not be backward compatible.
> 
> Could you explain? These changes fix wrong behavior of some of the endpoints that used to work with old versions of fastapi, but don't work with latest ones. There is no change in the API schema.
> My goal was to merge it so you can use it in your MR with tags so that they can be released together.

oh, if this is perfectly backward compatible, than sure. I was afraid that client from newest version would mismatch with apprunning on version 0.14 or lower.

rfhub2/db/model/keyword.py
pbylicki: json.loads/json.dumps will be enough for now

pbylicki:
@MaciejWiczk One thing, that came to my mind is that you could want to have this released before tags, so users can use this changes removing fastapi version limitation without using the new tags functionality. But I don't think that's really an issue

MaciejWiczk (commented):
Yes, I agree that this can be released as is, before tags. But without bumping version this will not go to pypi.

pbylicki:
@MaciejWiczk Sure, just wanted to get your thoughts. Bumped to 0.15

PR #210, opened 2020-04-04
Add displaying tags as chips on UI

pbylicki:


MaciejWiczk (approved):
Looking perfectly! :)

PR #211, opened 2020-04-05
Fix unit tests

pbylicki:


MaciejWiczk (approved):
Splendid, big thanks for help!

PR #215, opened 2020-04-09
Bugfix/rfhub2 214 fix missing alembic ini

MaciejWiczk:
This is a version that works locally on Windows and is able to be run from any directory. In v0.16-17 it run ok from rfhub2 directory, but any other gave error about missing alembic.ini file. I have no idea what happens on Travis, but it's totally different then my local env.
@pbylicki could You check on your local environment what's happening?

### DIFF
--- a/rfhub2/db/migrate.py
+++ b/rfhub2/db/migrate.py
@@ -3,6 +3,8 @@ from alembic.runtime.migration import MigrationContext
 from alembic import command
 from sqlalchemy.engine import Connection, Engine
 from sqlalchemy.engine.reflection import Inspector
+from pathlib import Path
+from os import chdir
 
 
 def has_tables(connection: Connection) -> bool:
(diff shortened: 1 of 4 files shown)


rfhub2/db/migrate.py
```diff
@@ -3,6 +3,8 @@
 from alembic import command
 from sqlalchemy.engine import Connection, Engine
 from sqlalchemy.engine.reflection import Inspector
+from pathlib import Path
+from os import chdir, getcwd
```
MaciejWiczk, line 7: I wanted to avoid it

rfhub2/db/migrate.py
```diff
@@ -18,10 +20,15 @@ def has_revision(connection: Connection) -> bool:
 
 
 def migrate_db(engine: Engine) -> None:
-    alembic_cfg = Config("rfhub2/alembic.ini")
+    alembic_cfg_path = Path(__file__).resolve().parent.parent.parent / "rfhub2" / "alembic.ini"
+    alembic_cfg = Config(alembic_cfg_path)
     with engine.begin() as connection:
+        cwd = getcwd()
+        chdir(Path(__file__).resolve().parent.parent / "alembic" / "versions")
```
MaciejWiczk, line 27: This had to be done so hat app can find alembic.ini from any directory it's invoked.

rfhub2/db/migrate.py
```diff
@@ -18,10 +20,15 @@ def has_revision(connection: Connection) -> bool:
...
         # check if database has not been migrated yet with alembic
         if has_tables(connection) and not has_revision(connection):
             # stamp database with the initial revision id, representing pre-alembic database schema
             command.stamp(alembic_cfg, "c54a916ec6c8")
         command.upgrade(alembic_cfg, "head")
+        chdir(cwd)
```
MaciejWiczk, line 33: I thought that it would be good to restore cwd

PR #226, opened 2020-05-02
Rfhub2 216 read the docs creation

### LINKED ISSUE #216: setup readthedocs for RfHub2 with extensive documentation
Apart from Readme.md and development.md it would be good to have searchable documentation about app. Autometed way of creating docs should be setup and updated each release.

MaciejWiczk:
Adding source of documentation in rst format, that will be used by sphinx and hosted on readthedocs. 
Closing #216


MaciejWiczk:
@pbylicki, there is a need to tag this git branch, before merging. On one hand, the app itself didn't change, so there is no need to bum version.py, but on the other hand, it's not exaclty the same as 0.18. What do You think I should do?
PS. Maybe we could extend CICD on master to tag github repository with the same version as in version.py? this way we would be safe, and docs will be published on readthedocs automatically?

docs/source/development.rst
```diff
@@ -0,0 +1,174 @@
...
+::
+
+    rfhub2-cli --no-installed-keywords ../your_repo ../your_other_repo
+
+Preserving already loaded collections and adding new ones:
+::
+
+    rfhub2-cli -l append ../your_repo ../your_other_repo
+
+Fronted
```
pbylicki: missing n in Frontend

docs/source/searching.rst
```diff
@@ -0,0 +1,31 @@
+Searching RfHub2
+----------------
+
+Browsing tables
+^^^^^^^^^^^^^^^
+All collected libraries and resoures are available on both on left-hand side menu and main table,
```
pbylicki: on both on => on both

rfhub2/templates/index.html
```diff
@@ -1 +1 @@
...
...1:0},c=[];function f(e){if(t[e])return t[e].exports;var r=t[e]={i:e,l:!1,exports:{}};return i[e].call(r.exports,r,r.exports,f),r.l=!0,r.exports}f.m=i,f.c=t,f.d=function(e,r,t){f.o(e,r)||Object.defineProperty(e,r,{enumerable:!0,get:t})},f.r=function(e){"undefined"!=typeof Symbol&&Symbol.toStringTag&&Object.defineProperty(e,Symbol.toStringTag,{value:"Module"}),Object.defineProperty(e,"__esModule",{value:!0})},f.t=function(r,e){if(1&e&&(r=f(r)),8&e)return r;if(4&e&&"object"==typeof r&&r&&r.__esModule)return r;var t=Object.create(null);if(f.r(t),Object.defineProperty(t,"default",{enumerable:!0,value:r}),2&e&&"string"!=typeof r)for(var n in r)f.d(t,n,function(e){return r[e]}.bind(null,n));return t},f.n=function(e){var r=e&&e.__esModule?function(){return e.default}:function(){return e};return f.d(r,"a",r),r},f.o=function(e,r){return Object.prototype.hasOwnProperty.call(e,r)},f.p="/static/";var r=window.webpackJsonpfrontend=window.webpackJsonpfrontend||[],n=r.push.bind(r);r.push=e,r=r.slice();for(var o=0;o<r.length;o++)e(r[o]);var s=n;a()}([])</script><script src="/static/static/js/2.16103137.chunk.js"></script><script src="/static/static/js/main.d7d8746e.chunk.js"></script></body></html>
```
pbylicki: to be reverted

tests/acceptance/cli_population.robot
```diff
@@ -27,13 +27,13 @@ Cli Should Populate App With Installed Keywords
     ...    XML library with 37 keywords loaded.
     ...    Easter library with 1 keywords loaded.
     ...    Process library with 15 keywords loaded.
-    ...    String library with 31 keywords loaded.
+    ...    String library with 32 keywords loaded.
```
pbylicki, line 30: These changes in tests are due to robot framework version update?

pbylicki:
> @pbylicki, there is a need to tag this git branch, before merging. On one hand, the app itself didn't change, so there is no need to bum version.py, but on the other hand, it's not exaclty the same as 0.18. What do You think I should do?
> PS. Maybe we could extend CICD on master to tag github repository with the same version as in version.py? this way we would be safe, and docs will be published on readthedocs automatically?

I think we need to tag commit on master once this branch is merged and deleted. We should use 0.18 as there is no change in the application itself and we can test the docs generation setup.

As for automatic tagging, we can look into this later, as current changes with CI are finalized.

tests/acceptance/cli_population.robot
MaciejWiczk: Yes, RFWK has been extended with new keyword and some changes to libdoc:
https://github.com/robotframework/robotframework/issues/3534

MaciejWiczk:
I messed up with tag 0.18, it's got pushed againts origin, not feature branch. Apparently, it's better to tag after merging to master. Is it fine by You? 

Edit:
readthedocs will produce latest even without git tag. Let's try that, and see if everything is alright

pbylicki:
> I messed up with tag 0.18, it's got pushed againts origin, not feature branch. Apparently, it's better to tag after merging to master. Is it fine by You?
> 
> Edit:
> readthedocs will produce latest even without git tag. Let's try that, and see if everything is alright

Yes, I don't think that any other approach will work.

PR #238, opened 2020-05-20
42 add animation loading

MarcinMaciaszek:
Please run it and check if it works as expected :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ install:
   - pip install -r requirements.txt -r requirements-dev.txt -r requirements-postgres.txt
   - pushd frontend; yarn install; popd
 before_script:
-  - wget "https://chromedriver.storage.googleapis.com/80.0.3987.106/chromedriver_linux64.zip"
+  - wget "https://chromedriver.storage.googleapis.com/83.0.4103.39/chromedriver_linux64.zip"
   - unzip chromedriver_linux64.zip
   - sudo mv chromedriver /usr/local/bin
   - "export DISPLAY=:99.0"
--- a/frontend/src/components/CircularLoading.tsx
+++ b/frontend/src/components/CircularLoading.tsx
@@ -1,32 +1,25 @@
 /* eslint-disable no-script-url */
-
 import React from 'react';
-import { observer } from 'mobx-react';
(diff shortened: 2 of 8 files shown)


MaciejWiczk:
Ok, I have tested in on corp. env where times for loading collections are a bit longer. Left navigation panel works as expected! :)
I can imagine, that there is one more case, where spinner could be useful: Under main table on the first page, where You have collection list, You can have left panel hidden, but when You scroll to bottom of page, more collections are loaded, the same way it works for left navigation panel. Could You apply it also there?

pbylicki:
@MarcinMaciaszek @MaciejWiczk One thing that I see as potential trouble is that spinners in three different views relying on three different requests to load the data are sharing single loading flag. That may probably cause race condition if more than one of the requests is triggered at the same time. Have you seen something like that on slow network connection @MaciejWiczk ? If it's not causing any weird effect because of that then it's fine.

MarcinMaciaszek:
@MaciejWiczk 
Will place it there ;-)

@pbylicki 
You're right. I found an issue ;)
Do you have any idea, how to check progress in a convenient way?

pbylicki:
@MarcinMaciaszek the easiest way would be to have separate flags for them. At this point our store is flat so adding multiple flags won't look that good and may be error prone with number of flags growing, but maybe in next wave of refactoring we could store all parts of store data in the container objects or nested stores that will include flag indicating if data loading is in progress. Then only nested stores would be passed to the relevant component and it would have access only to single loading flag in its scope.

MaciejWiczk:
> That may probably cause race condition if more than one of the requests is triggered at the same time. Have you seen something like that on slow network connection @MaciejWiczk ? If it's not causing any weird effect because of that then it's fine.

@pbylicki I haven't seen it, altough I have not used it extensively.

>  Will place it there ;-)

Cool, in future, when we add new aplication features, the left panel would change probably, to list all app feature (keyword, testcases, executions, dashboards etc.) so we will stay with one less circular progress :)

MaciejWiczk (approved):
That is exactly what was needed. Great job! :)

MaciejWiczk:
Last thing to do is to bump version.py, so that it would be released to pypi

pbylicki (approved):

PR #241, opened 2020-05-30
WIP: Feature/114 poc of testcase documentation

pbylicki:


rfhub2/cli/test_cases/test_cases_extractor.py
```diff
@@ -0,0 +1,121 @@
...
+        Creates TestDoc objects based on robot.testdoc.TestSuiteFactory method.
+        """
+        return [suite for path in self.paths for suite in self.create_testdoc_from_path(path)]
+
+    def create_testdoc_from_path(self, path: Path):
+        """
+        Creates TestDoc object based on robot.testdoc.TestSuiteFactory method.
+        """
+        testdoc = TestSuiteFactory(str(path))
+        return [self.serialize_testdoc(testdoc)] + self._traverse_suites(testdoc)
```
pbylicki: It looks like this first serialize_testdoc call could be part of _traverse_suites method if it would first serialize_testdoc of top level suite and then for each child suite call traverse_suites and combine results

rfhub2/cli/test_cases/test_cases_extractor.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        Returns list of tests.
+        """
+        return [Test(doc=test.doc, id=test.id, longname=test.longname, name=test.name, parent=test.parent.name,
+                     source=test.source if hasattr(test, 'source') else None,
+                     tags=list(test.tags), template=test.template, timeout=test.timeout,
+                     keywords=self.get_normal_keywords(test), setup=self.get_setup_keywords(test),
+                     teardown=self.get_teardown_keywords(test)) for test in testdoc.tests]
+
+    def get_setup_keywords(self, testdoc: TestSuite) -> List[Keyword]:
```
pbylicki: These 3 methods could be turned into single method that accepts Union[List[Keyword], Keyword, None]

rfhub2/cli/test_cases/test_cases_extractor.py
MaciejWiczk: Wouldn't that end in duplicating suite objects? If top level serialization would be added at the beginning, it would treat sub-suite as top level suite for every method call. Or am I wrong? My force is not that strong when recurrence calls.

rfhub2/cli/test_cases/test_cases_extractor.py
MaciejWiczk: Yeah, I just thought of using has.attr, that would be used with method parameter, like normal, teardown, setup. I will add some changes tomorrow

rfhub2/cli/test_cases/test_cases_extractor.py
MaciejWiczk: I figured out what You meant. It works! :)

rfhub2/cli/test_cases/test_cases_extractor.py
MaciejWiczk: I did some changes for this one, but I don't think it looks good. I believe I didn;t get what You had in mind. Anyway, I merged setup/teardown keyword into one method with kw_type parameter, and simplified get_normal_keywords method into single line list comprehension.

pbylicki:
@MaciejWiczk server-side changes were merged including model definition. It should be re-used on client side.

PR #295, opened 2020-11-15
Hotfix/fix building ui

MaciejWiczk:
Yarn version changed on travis CI, and didn't work with actual node_modules, making CI fail. I tired other solutions, like downgrading yarn on travis, but this is not as straightforward as I thought. So in the end there is a brute force approach, to install all of the packages during CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,9 @@ before_install:
 #  - echo "deb http://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
   - sudo apt-get update -qq
 #  - sudo apt remove -y yarn
+  - yarn --version
   - wget https://github.com/yarnpkg/yarn/releases/download/v1.19.1/yarn_1.19.1_all.deb && sudo dpkg -i yarn_1.19.1_all.deb
+  - yarn --version
 install:
   - pip install -r requirements.txt -r requirements-dev.txt -r requirements-postgres.txt
   - pushd frontend; yarn install; popd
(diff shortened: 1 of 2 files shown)


pbylicki:
@MaciejWiczk Did you try to just remove caching node_modules? It is defined in this section:
`cache:
  pip: true
  yarn: true
  directories:
    - frontend/node_modules`

Maybe we could later re-enable it once it starts to build using right dependencies.

MaciejWiczk:
> @MaciejWiczk Did you try to just remove caching node_modules? It is defined in this section:
> `cache: pip: true yarn: true directories: - frontend/node_modules`
> 
> Maybe we could later re-enable it once it starts to build using right dependencies.

Yes, You are right, I will revert my changes to build_ui.sh script, and remove caching from travis

.travis.yml
```diff
@@ -11,12 +11,10 @@ cache:
   pip: true
   yarn: true
   directories:
-    - frontend/node_modules
 before_install:
   - sudo apt-key adv --fetch-keys http://dl.yarnpkg.com/debian/pubkey.gpg
   - echo "deb http://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
   - sudo apt-get update -qq
-  - sudo apt-get install -y -qq yarn
```
pbylicki, line 19: if we no longer need to install yarn then this whole before_install section is not needed

MaciejWiczk: > if we no longer need to install yarn then this whole before_install section is not needed

done :)

pbylicki (approved):

PR #346, opened 2021-02-23
Refactor statistics extractor to use Robot API visitor

bhirsz:
It's proposal of refactor for extracting keywords from result using Robot parsing API visitors.

It should be tiny bit faster on larger files (150Mb+). I did not test it thorougly yet - it's draft.

### DIFF
--- a/rfhub2/cli/statistics/statistics_extractor.py
+++ b/rfhub2/cli/statistics/statistics_extractor.py
@@ -1,129 +1,40 @@
-from collections import Counter
-from dataclasses import dataclass
+from collections import defaultdict
 from datetime import datetime
-from itertools import groupby
-from typing import Callable, Dict, Iterable, List, Tuple
-from xml.etree.ElementTree import Element, parse
+from typing import List
+
+from robot.api import ExecutionResult, ResultVisitor
 
 from rfhub2.model import KeywordStatistics
 
 
-@dataclass
-class XmlKeyword:
-    library: str
-    name: str
-    elapsed: int
+class KeywordsCollector(ResultVisitor):
+    def __init__(self):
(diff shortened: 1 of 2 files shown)


pbylicki:
Thanks @bhirsz, it looks good! If you could also add tests for that it would be great.

MaciejWiczk:
@bhirsz I was looking into that one, and one thing is bottering me. Which robotframework version is intended to use with this feature? 3.2.2 seems to be not enough:
```
File "C:\repo\rfhub2\rfhub2\cli\statistics\statistics_extractor.py", line 15, in visit_keyword
    keyword.body.visit(self)
AttributeError: 'Keyword' object has no attribute 'body'
```
If this feature needs RFWK 4.0 I will mark it appropriate Milestone.

bhirsz:
@MaciejWiczk I apologise for leaving this PR - I was bit busy recently. It should support 3.2 with one change - in RF4 "keywords" attribute was renamed to body. So depending on version you should access it with different name

bhirsz:
But when I think about it its possible we could replace that line with self.generic_visit (and it should work for both 3.2 and 4)

MaciejWiczk:
>I apologise for leaving this PR - I was bit busy recently.

You've been busy, with great results! :)

> But when I think about it its possible we could replace that line with self.generic_visit (and it should work for both 3.2 and 4)

Oh cool, I will take a look at this approach and see what's what :)

MaciejWiczk:
@bhirsz is this what You mean:
```
    def visit_keyword(self, keyword):
        try:
            keyword.body.visit(self)
        except AttributeError:
            keyword.keywords.visit(self)
        self.statistics[keyword.name].append(keyword)
```

PR #355, opened 2021-03-19
[rfhub-342] Added importing installed libraries by their names

### LINKED ISSUE #342: add possibility to import installed libraries using name instead of full path
Currently when one wants to import python package that is installed, have to provide full package installation directory, using ie:
```
python -c "from distutils.sysconfig import get_python_lib; print(get_python_lib()+'\SeleniumLibrary')"
```
This could be enhanced to use only module name, and let rfhub do the rest of the work.

MaciejWiczk:
PR to implement #342.
I'm not sure if using name `paths` should be kept in cli.py and so on. Maybe libraries/collections would be a better suiting one?

### DIFF
--- a/rfhub2/cli/keywords/keywords_extractor.py
+++ b/rfhub2/cli/keywords/keywords_extractor.py
@@ -34,7 +34,9 @@ class CollectionUpdateWithKeywords:
 
 
 class KeywordsExtractor:
-    def __init__(self, paths: Tuple[Union[Path, str], ...], no_installed_keywords: bool) -> None:
+    def __init__(
+        self, paths: Tuple[Union[Path, str], ...], no_installed_keywords: bool
+    ) -> None:
         self.paths = paths
         self.no_installed_keywords = no_installed_keywords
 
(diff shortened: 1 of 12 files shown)


rfhub2/cli/keywords/keywords_extractor.py
```diff
@@ -45,12 +48,30 @@ def get_libraries_paths(self) -> Set[Path]:
         """
         libraries_paths = set()
         for path in self.paths:
-            libraries_paths.update(self._traverse_paths(Path(path)))
+            path = self.get_library_path_from_name(path)
+            if path:
```
MaciejWiczk, line 52: @MarcinMaciaszek  You are right that this is ugly. The other option would be to move the verification of path to _traverse_paths, but 
[that one is truly ugly as hell.](https://github.com/pbylicki/rfhub2/blob/master/rfhub2/cli/keywords/keywords_extractor.py#L54)
Other options that I can see is to use try/except or continue, but would they be better?

PR #365, opened 2021-03-29
simple UI modification & app helm chart

justyna363:
added a name&color to drawer to be similar like appBar

image


pbylicki:
Hi @justyna363, could you add a section in Readme file describing how to install this helm chart?

justyna363:
added :)

Issue #55, opened 2019-09-17
Rfhub2 does not support other docformats than robot


Snooz82:
Rfhub is nice!

It look like it renders the robot docfomat to html.
But if there is a library that has rst as doc format it just shows the plain source code.

In the moment the only way to get rendered html from Robot Framework is, parsing the html libdoc output.

I made a pull request to RobotFramework that the xml output can contain html as doc.
Independent of the original library doc format, the xml will always has ready made html.

When this PR is accepted, you should use Libdoc with `--format XML:HTML`

Regards
René

MaciejWiczk:
@Snooz82 I knew Your face was familiar, and now I know where I've seen You :)
That would be good enhancement to have that in rfhub2 as well. Hopefully RFWK 3.2 with new parser arrives soon.

Snooz82:
Where have you seen me?

MaciejWiczk:
> Where have you seen me?

Well, your avatar, on rfwk LibDoc PR. I keep an eye on that repo ;)

MaciejWiczk:
@Snooz82 I've taken a look at feature, and here are my thoughts:
- 'XML:HTML' format is aplicable for generating libdoc output XML files, that contains HTML in nodes itself, currently RfHub does not handle this, and importing such file ends up in having raw HTML displayed. Paragraph ends up inside another paragraph and quoted. This will only happen when loading lidoc.xml file from library, while my idea was to just scan all paths and generate LibDoc object, that is later on inserted into DB via API.
- having library documentation with hiperlinks inside docs like in [here](https://robotframework.org/robotframework/latest/libraries/BuiltIn.html) would definetely be nice and useful for users. I think that there are two ways of adding this: either to add hyperlink to specific keyword based on RfHub keyword ID. like on left tab, or to name each and every paragraph with keyword with it's name as ID, like in original Libdocs.

@pbylicki what would You advise me?

Ka55i0peia:
After diving into source code and db structure, I've found this hard code line beeing the root cause of this limitation:

https://github.com/pbylicki/rfhub2/blob/d4e9b5bbe4d789f62a1acf302c121dbba8f5639f/rfhub2/db/model/mixins.py#L16

**Edit**

Its a bit of hackity hack reverse engineered / trail and error debugging, but this patch might goes into the right direction ...

```
    class DocMixin:
    #...
    @property
    def html_doc(self) -> str:
        try:
[...]

PR #371, opened 2021-04-06
Add support for name version doc format libdoc

krysalead:
This PR add the support of the following libdoc parameters
- Name: using the --name you can override the name of the library
- version: using --version you can specify the version 
- doc-format: using the --doc-format you can specify the format of the doc.

Sorry if the code is not following the philosophy. 
I have also fixed the tests to work with Robot4.0

Regards

K

### DIFF
--- a/rfhub2/cli/cli.py
+++ b/rfhub2/cli/cli.py
@@ -44,6 +44,13 @@ from rfhub2.cli.statistics.statistics_importer import StatisticsImporter
     help="Specifies name of the library you are pushing the collection",
 )
 @click.option(
+    "-d",
+    "--doc-format",
+    type=click.STRING,
+    default=None,
+    help="Specifies format of the documentation libdoc will analyse",
+)
+@click.option(
     "--no-installed-keywords",
     type=click.BOOL,
     default=False,
@@ -83,13 +90,15 @@ def main(
     no_installed_keywords: bool,
     version: str,
     name: str,
+    doc_format: str
+
 ) -> None:
     """Package to populate rfhub2 with robot framework keywords
        from libraries and resource files."""
     client = Client(app_url, user, password)
     if mode == "keywords":
         rfhub_importer = KeywordsImporter(
-            client, paths, no_installed_keywords, load_mode, name, version
+            client, paths, no_installed_keywords, load_mode, name, version, doc_format
         )
         loaded_collections, loaded_keywords = rfhub_importer.import_data()
         print(
--- a/rfhub2/cli/cli.py
+++ b/rfhub2/cli/cli.py
@@ -30,6 +30,20 @@ from rfhub2.cli.statistics.statistics_importer import StatisticsImporter
     help="Specifies rfhub2 password to authenticate on endpoints that requires that. Default value is rfhub.",
 )
 @click.option(
+    "-v",
+    "--version",
+    type=click.STRING,
+    default=None,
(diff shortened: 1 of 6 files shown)


MaciejWiczk:
Hi @krysalead, thanks for PR! Could I ask You to walk me through the situation, when You would like to load more than one library? Will all of them be overwritten by name, version and doc-format?

Ka55i0peia:
@krysalead . Please note the relation to issue https://github.com/pbylicki/rfhub2/issues/55 .

I think the doc format _might not be set explicit on cli_. `robot.libdoc` is extracting it dynamic from resource it is going to read. (`robot`, if docing a robot resource file, or [if specified in python](https://github.com/robotframework/robotframework/blob/master/doc/userguide/src/SupportingTools/Libdoc.rst#documentation-syntax) using that format).

Back from a short holiday I'll try to push my employer to make suggestions/investigation I made in https://github.com/pbylicki/rfhub2/issues/55 public as pull request.

krysalead:
Hi @MaciejWiczk there is multiple issues
1. There is no way to provide a version when the keyword are inside a robot file.
2. There is no way to easily link all the keywords of a library in the context of multi library documentation. This is not the best solution but an acceptable quick win. Ideally having "folders" will be quite interesting, also having the information about the name of the lib to be installed and then the reference to the path of the file to put in the "Settings" section of the robot test will be the perfect solution.
3. The format of the documentation we are using in Robot file seems to not being well rendered in the rfhub, due to the default format used by the testdoc lib.

So the goal of this PR is to allow an override of those information from the command line, if not it will still use the one detected.
HI @Ka55i0peia, weird that libdoc is not properly detecting the format, carriage return are not kept and my doc is on a single line. My change is fixing that, maybe something I missed in my documentation.

krysalead:
image
This is an example of my lib that contains files which are in the end inside multiple collection, so what I was looking for is a top level object called lets say **library** to group them
image

MaciejWiczk:
Hi @krysalead let me go through things You mentioned below
> Hi @MaciejWiczk there is multiple issues
> 
> 1. There is no way to provide a version when the keyword are inside a robot file.

Yes, but that was default with LibDoc AFAIK

> 2. There is no way to easily link all the keywords of a library in the context of multi library documentation. This is not the best solution but an acceptable quick win. Ideally having "folders" will be quite interesting, also having the information about the name of the lib to be installed and then the reference to the path of the file to put in the "Settings" section of the robot test will be the perfect solution.

Combinig robot files in python library is a new approach for me. I have not tought about such a solution. And oribably LibDoc is handling that the same way that RfHub2 does. A workaround I can think of is to create LibDoc file and use it as a source for RfHub2 with version, name etc provided from LibDoc cli.

> 3. The format of the documentation we are using in Robot file seems to not being well rendered in the rfhub, due to the default format used by the testdoc lib.

I know, and hopefully things will be adressed in future

 
> So the goal of this PR is to allow an override of those information from the command line, if not it will still use the one detected.

This must be handled with more attention. RfHub2-cli was created mainly for batch mode and to relieve user from thinking. Your changes are targeted for single library in mind.
 
> HI @Ka55i0peia, weird that libdoc is not properly detecting the format, carriage return are not kept and my doc is on a single line. My change is fixing that, maybe something I missed in my documentation.

Could You show me the line with changes? I just got back form time off and I'am still a bit puzzled

PR #372, opened 2021-04-10
Add g++ to fix Docker Alpine builds

pbylicki:
Recent update of SQLAlchemy version to 1.4.x introduced new dependency - greenlet which requires c++ compiler to build during Alpine Docker build.


MaciejWiczk:
You were too fast for me to review ;)

Issue #314, opened 2020-12-13
Add an endpoint to remove all collections from rfhub2


MaciejWiczk:
It would speed up a bit of collections removal if done by one request, instead of multiple, one by one, requests. As a bonus, code [here](https://github.com/pbylicki/rfhub2/blob/master/rfhub2/cli/keywords/keywords_importer.py#L35) could be reduced a bit.

PR #381, opened 2021-04-26
Add delete all collections endpoint and udpate CLI (#314)

MarcinMaciaszek:


rfhub2/cli/keywords/keywords_importer.py
```diff
@@ -48,8 +48,7 @@ def _delete_collections(self) -> Set[int]:
         """
         collections = self.client.get_collections()
         collections_id = {collection["id"] for collection in collections}
-        for id in collections_id:
-            self.client.delete_collection(id)
+        self.client.delete_all_collection()
```
MaciejWiczk: I think that this method might not be needed anymore at all. All it was doing is done by `delete_all_collections` method

MarcinMaciaszek: true

rfhub2/cli/keywords/keywords_importer.py
```diff
@@ -32,26 +32,17 @@ def delete_all_collections(self) -> Set[int]:
         """
```
MaciejWiczk: This whole method is not needed, due to having 
```
self.client.delete_all_collections()
```

MarcinMaciaszek: I have removed it, it should work well still. The only thing is the coverage. Have no idea how to cover line: 110 in collections.py
@MaciejWiczk do you have any idea how to handle it? or just remove this else?

MaciejWiczk: Probably it's safe to go without else. About coverage, don't try to cover everything for sake of it :)

MaciejWiczk (approved):

PR #412, opened 2021-06-13
Auto update chromedriver version in Travis Job

Vincema:
Since most of the CI builds are failing due to the incompatibility chrome <-> chromedriver, this PR adds a slight change to the travis job config: now, the chromedriver is installed matching the currently installed chrome version. It removes the need to manually bump the chromedriver version each time a new chrome update is released.

The script follows the procedure given here: https://chromedriver.chromium.org/downloads/version-selection.

(BTW, https://travis-ci.org will be closed down for good on June 15th so you may migrate to https://travis-ci).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,10 @@ install:
   - pip install -r requirements.txt -r requirements-dev.txt -r requirements-postgres.txt
   - pushd frontend; yarn install; popd
 before_script:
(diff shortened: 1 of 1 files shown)


MaciejWiczk:
I really like this PR! :) @pbylicki would You have some time to switch to travis-ci form travis.org?

PR #415, opened 2021-06-14
Add new options to filter keywords by their tag

### LINKED ISSUE #361: RFHub to use tags for filtering like robot
Hi,

Some of the keyword I am using are not to be used directly, it would be nice to have a way to filter for instance using the tags we can put in the robot file. It won't work with python and maybe this could be done with also a filtering on python script. 
WDYT?

Vincema:
Hi all!

This PR brings a new feature: filtering keywords by their tags (fixes #361).

`rfhub-cli` now has 2 more options: `--include` and `--exclude`. They behave the same way as the  `--include` and `--exclude` options from RobotFramework (see [robot user manual](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#tag-patterns)).

Few examples:
```bash
# Includes only keywords containing "tag1"
rfhub2-cli --include tag1 .

# Includes keywords with tags "tag1" or "tag2" and exclude keywords with tag "tag3" and "tag4" but not "tag5"
rfhub2-cli --include tag1ORtag2 --exclude tag3ANDtag4NOTtag5 .

# Includes keywords with tags starting with "awesome-tag-"
rfhub2-cli --include awesome-tag-* .

# Tag pattern are underscore and case insensitive:
#   Includes tag "my-tag":
rfhub2-cli --include _?_y-TA____g .
```

I have however a few points I'm not really sure about:

- WDYT of the help sections and readme about this new feature? Are they understandable and clear enought? (I'm not native so please feel free to correct them)
- from now on, if a user includes or excludes a tag, it might find himself with a lot of loaded but empty collections (since all the keywords were filtered). Even thought it's not a big matter, is there anything that can be done about that? 
- should we add the short options "-i" and "-e"? (currently not implemented)

Note also that this feature is not implemented for statistics mode.

Anyway let me now what you think!

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,13 +74,24 @@ To populate app but to skip loading RFWK installed libraries:
 ```
 rfhub2-cli --no-installed-keywords ../your_repo ../your_other_repo
 ```
-##### Rfhub2-cli for keywords documentation can be run in four load-modes:
+To populate app only including or excluding keywords with tags matching custom patterns (more information on the include/exclude feature [here](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#tag-patterns)):
+```bash
+# Including keywords containing tag "your-tag" and not containing tag "your-other-tag":
+rfhub2-cli --include your-tag --exclude your-other-tag ../your_repo ../your_other_repo
 
+# To populate app only with keywords containing either tag "💡" or "🔧", and tag "🤖":
+rfhub2-cli --include 💡OR🔧AND🤖 ../your_repo ../your_other_repo
+
+# To populate app only with keywords containing tags starting by "important-tag-":
+rfhub2-cli --include important-tag-* ../your_repo ../your_other_repo
+```
+
+##### Rfhub2-cli for keywords documentation can be run in three load-modes:
 - `merge`, default mode, updates only matched collections, does nothing with not matched ones
 ``` rfhub2-cli --load-mode=merge ../your_repo ../your_other_repo```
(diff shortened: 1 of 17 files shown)


MaciejWiczk:
@Vincema sorry for late reply. I think this is a good idea to follow robot CLI with tags behaviour. 

>should we add the short options "-i" and "-e"? (currently not implemented)

I think it would be a good idea to do that.

>WDYT of the help sections and readme about this new feature? Are they understandable and clear enought? (I'm not native so please feel free to correct them)

I don't see anything to correct. I wouldn't put it any better myself :)

> from now on, if a user includes or excludes a tag, it might find himself with a lot of loaded but empty collections (since all the keywords were filtered). Even thought it's not a big matter, is there anything that can be done about that?

This might be a surprise for someone if their keywords are not visible, especially when working in distributed company. Disclaimer/info for each library on frontend would be helpful, saying that this collection was loaded with include/exclude option, but that would require DB and model changes. I guess it is waaaay too much for now.

>Note also that this feature is not implemented for statistics mode.

I guess that putting an issue to implement that in statistics would be enough :)

Were You able to run travis CI on your repo clone? Sadly, they are not run on PR from other branches :(

Vincema:
Hi! 
Thanks for reviewing this PR!
I've added the short options -i/-e.
The CI checks are all green on my clone so we should be good to go 👍

MaciejWiczk:
>The CI checks are all green on my clone so we should be good to go

@Vincema let's merge it! :)

MaciejWiczk (approved):

Issue #291, opened 2020-10-23
Support variable in keyword name for statistics


leeuwe:
Currently, it looks to me as if keywords using the Embedded arguments approach are not counted in the Statistics. For me, that's a dealbreaker. I have little knowlede about how this should work, but I think it can work if the report.xml file is create with TRACE logging.
This would give the following:
<kw name="I click on activity &quot;xyz&quot;" library="GebiedsInfo">
<msg timestamp="20201022 13:05:45.004" level="TRACE">Arguments: [ ${activity}='xyz' ]</msg>

With this one should be able to reverse engineer what the keyword with embedded argument was.

MaciejWiczk:
Hi @leeuwe, You are right that embedded keywords are not taken into account when calculating statistics. The root of this issue lies between RfHub and Robotframework itself. When parsing execution results, to collect statistics, RfHub ges through output.xml file, that looks like below:
```
<kw name="Open ${single_class_lib} In Left Panel">
<kw name="Run Keyword And Return Status" library="BuiltIn">
<doc>Runs the given keyword with given arguments and returns the status as a Boolean value.</doc>
<arguments>
<arg>Element Should Not Be Visible</arg>
<arg>${hamburger}</arg>
```
At first line You have embedded keyword, however it names differ from its definition `Open ${library} In Left Panel`, because of different variable used in test. If string value is used instead of variable, it works the same. Matching definition with execution is a tricky one to do. I was thinking of rising issue on RFWK side, so thath original keyword definition would be saved in output.xml, so that root cause would be fixed. 

Currently for joining those data on RfHub side we use SQL joins. I think this could be updated for embedded keywords, to remove variables from them, but that would be half-good solution, because of injecting literall instead of variable in keyword execution. I;m going to investigate regexp_join but that's wrong from perfomance point of view.

leeuwe:
Thanks for your expanation. This is why I suggested this should be possible when one logs with TRACE, as this would make it doable in my opinion:

```
<test id="s1-s1-s1-t1" name="Test Embedded Logging">
<kw name="Userkeyword with ${Override}">
<msg timestamp="20201026 10:55:10.354" level="TRACE">Arguments: [ ${EmbeddedArgument}='https://www.google.com' ]</msg>
<kw name="Open Browser" library="SeleniumLibrary">
<doc>Opens a new browser instance to the optional ``url``.</doc>
<arguments>
<arg>${EmbeddedArgument}</arg>
<arg>gc</arg>
[...]

PR #440, opened 2021-09-12
Update rfhub2 to handle rfwk 4.0+

MaciejWiczk:
Hopefuly, this will enable RfHub2 to support RFWK 4.0+. I do not have an acess to large RFWK repository anymore, so maybe someone could test that locally, before releasing?
@MarcinMaciaszek ?

PR #443, opened 2021-11-13
Embedded keywords statistics are now extracted and displayed

MaciejWiczk:
PR with changed to extract embedded keywords statistics 
#291


rfhub2/version.py
```diff
@@ -1 +1 @@
-version = "0.29"
+version = "0.30"
```
pbylicki, line 1: If you are going to merge it to #440 then there's probably no need to update version

pbylicki (approved):

rfhub2/version.py
MaciejWiczk: Thanks @pbylicki I wanted to keep the merge order and introduce smaller changes at once. After #440 would be merged, I will change base branch to master :)

PR #449, opened 2021-12-03
[TestCases] Added Get Suites

MaciejWiczk:
@pbylicki I'm trying to figure out how to add suites with testcases, but I would need your help on that. I did some WIP code in this PR, but I'm missing logic on hierarchies etc. Could You navigate me?


pbylicki:
> @pbylicki I'm trying to figure out how to add suites with testcases, but I would need your help on that. I did some WIP code in this PR, but I'm missing logic on hierarchies etc. Could You navigate me?

@MaciejWiczk The main idea there, was that we only add/delete suites as suite hierarchies, not individual suites. That means that endpoint should expect a full suite hierarchy, which is a tree of nested suites, with the tree root being top-level suite, not a part of other suite. It simplifies creation of individual suites and relations between them so that we have easy read access to them and can treat them as regular resource then. Once the suites are saved and you have the IDs returned, adding test cases is straightforward because each test case is linked with specific suite.
So these endpoints should just wrap existing repository methods, with no special logic involved, it's already handled there. There will be more work on CLI side to build the hierarchies from provided suites

rfhub2/api/endpoints/test_cases.py
```diff
@@ -0,0 +1,99 @@
...
+    repository: TestCaseRepository = Depends(get_test_case_repository),
+    suite_repository: SuiteRepository = Depends(get_suite_repository),
+    test_case: TestCaseCreate,
+):
+    suite: Optional[DBSuite] = suite_repository.get(test_case.suite_id)
+    if not suite:
+        raise HTTPException(status_code=400, detail="Suite does not exist")
+    try:
+        db_test_case: DBTestCase = repository.add(DBTestCase.create(test_case))
+        return repository.get(db_test_case.id)
```
MaciejWiczk, line 70: This is wrong, and I know I'm missing something. @pbylicki could You take a look?

pbylicki: Can you tell what is the error you're facing? I don't see any unit tests for that yet and just by looking at this code it seems to be doing the right thing (map payload to db model -> create record and return it with id -> get full record including suite data by that id)

rfhub2/api/endpoints/suites.py
```diff
@@ -0,0 +1,81 @@
...
+        use_tags=use_tags,
+        skip=skip,
+        limit=limit,
+        ordering=ordering,
+    )
+
+
+@router.get("/{id}/", response_model=Suite)
+def get_suite(*, repository: SuiteRepository = Depends(get_suite_repository), id: int):
+    suite: Optional[DBSuite] = repository.get(id)
```
pbylicki, line 43: this is not DBSuite, repository already maps it to response model

rfhub2/api/endpoints/test_cases.py
```diff
@@ -0,0 +1,99 @@
...
+        limit=limit,
+        ordering=ordering,
+    )
+
+
+@router.get("/{id}/", response_model=TestCase)
+def get_test_case(
+    *, repository: TestCaseRepository = Depends(get_test_case_repository), id: int
+):
+    testcase: Optional[DBTestCase] = repository.get(id)
```
pbylicki, line 53: this is not DBTestCase, repository already maps it to response model

rfhub2/api/endpoints/test_cases.py
```diff
@@ -0,0 +1,99 @@
...
+
+@router.post("/", response_model=TestCase, status_code=201)
+def create_test_case(
+    *,
+    _: bool = Depends(is_authenticated),
+    repository: TestCaseRepository = Depends(get_test_case_repository),
+    suite_repository: SuiteRepository = Depends(get_suite_repository),
+    test_case: TestCaseCreate,
+):
+    suite: Optional[DBSuite] = suite_repository.get(test_case.suite_id)
```
pbylicki, line 65: this is not DBSuite, repository already maps it to response model

rfhub2/api/endpoints/suites.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@router.delete("/")
+def delete_all_suites(
+    *,
+    _: bool = Depends(is_authenticated),
+    repository: SuiteRepository = Depends(get_suite_repository),
+):
+    deleted: int = repository.delete_all_hierarchies()
+    if deleted:
```
pbylicki, line 78: IMO these delete all endpoints for suites and test cases should return response including number of deleted records, like we do for Keyword Statistics (and also skip the case of returning 404 for 0 rows found).
I know that we have it implemented like this for Collections and I missed it before, but I think my proposal gives more meaningful output that can be also included in CLI logs for visibility.

rfhub2/cli/test_cases/td.py
```diff
@@ -0,0 +1,8 @@
+from robot.testdoc import TestSuiteFactory
```
pbylicki, line 1: can be deleted

rfhub2/api/endpoints/test_cases.py
```diff
@@ -0,0 +1,99 @@
...
+):
+    deleted: int = repository.delete(id)
+    if deleted:
+        return Response(status_code=204)
+    else:
+        raise HTTPException(status_code=404)
+
+
+@router.delete("/")
+def delete_all_test_cases(
```
pbylicki, line 90: is this endpoint needed? When you delete the suites (and I see it is used in CLI), it also deletes related test cases.

PR #451, opened 2021-12-07
Add sorting CollectionWithStats by keyword count

pbylicki:
Based on PR #450 with a goal to simplify the changes.
First commit adds calculated column `keyword_count` to CollectionWithStats model and enabled ordering by this column. It also enables it for existing calculated column `times_used`. Both calculated fields are updated to non-optional returning 0 when no keywords or usage found.
Second commit adds fixes for SqlAlchemy deprecation warnings present since version 1.4.

Version is not updated on purpose to release it once UI changes are ready.


MaciejWiczk (approved):
Ok, now I get it :) Guess we will close #450.

